PR #3, opened 2021-03-10
Use interface for messaging and commands

muneebahmed10:
*Description*:
Removes the functions `Agent_MessageSend`, `Agent_MessageReceive`, `Agent_GetCommand`, and `Agent_ReleaseCommand`, and replaces them with a struct of function pointers similar to MQTT's transport interface.


source/mqtt_agent.c
```diff
@@ -732,7 +735,7 @@ MQTTStatus_t MQTTAgent_Init( MQTTAgentContext_t * pMqttAgentContext,
         {
             pMqttAgentContext->pIncomingCallback = incomingCallback;
             pMqttAgentContext->pIncomingCallbackContext = pIncomingPacketContext;
-            pMqttAgentContext->pMessageCtx = pMsgCtx;
+            pMqttAgentContext->agentInterface = *pMsgInterface;
```
leegeth, line 738: I think it is better to validate the members of this structure also here.

leegeth (approved):

leegeth (approved):

dan4thewin (approved):

PR #4, opened 2021-03-10
Decrease resume session complexity

muneebahmed10:
*Description*:
- Refactor Resume session to decrease complexity
- Update MQTTAgent_ProcessLoop

### DIFF
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -200,6 +200,24 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
                                          uint32_t blockTimeMs );
 
 /**
+ * @brief Resend QoS 1 and 2 publishes after resuming a session.
+ *
+ * @param[in] pMqttAgentContext Agent context for the MQTT connection.
+ *
+ * @return MQTTSuccess if all publishes resent successfully, else error code
+ * from #MQTT_Publish.
+ */
+static MQTTStatus_t resendPublishes( MQTTAgentContext_t * pMqttAgentContext );
+
+/**
+ * @brief Clear the list of pending acknowledgments by invoking each callback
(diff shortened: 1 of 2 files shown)


leegeth (approved):

source/mqtt_agent.c
```diff
@@ -200,6 +200,24 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
...
+ * @param[in] pMqttAgentContext Agent context for the MQTT connection.
+ *
+ * @return MQTTSuccess if all publishes resent successfully, else error code
+ * from #MQTT_Publish.
+ */
+static MQTTStatus_t resendPublishes( MQTTAgentContext_t * pMqttAgentContext );
+
+/**
+ * @brief Clear the list of pending acknowledgments by invoking each callback
+ * with #MQTTBadParameter.
```
dan4thewin, line 215: Using MQTTBadParameter strikes me as odd.  Would MQTTRecvFailed be a better choice?

muneebahmed10: Good catch. I actually used MQTTBadResponse but documented it incorrectly. Should I change them both to MQTTRecvFailed or keep as MQTTBadResponse?

dan4thewin: I don't think there is a best answer.  My first thought for bad response would be "I got something and it was bogus."  The receive failed makes me think "I called socket read and got an error." which is still not quite right, but is a better fit for a lost connection.  What do you think?

dan4thewin (approved):

source/mqtt_agent.c
muneebahmed10: I think best would be if this situation had its own error code, but given the current implementation decision of using coreMQTT's codes, I agree with you that receive failed is the closest fit.

PR #5, opened 2021-03-18
CBMC proofs for MQTTAgent_Init

leegeth:
This PR contains
1. CBMC proof for MQTTAgent_Init API
2. Common infrastructure support for writing CBMC proofs for coreMQTT-Agent APIs.


test/cbmc/sources/mqtt_agent_cbmc_state.c
```diff
@@ -0,0 +1,36 @@
...
+ */
+#include <stdint.h>
+#include <stdlib.h>
+#include "mqtt_agent_cbmc_state.h"
+
+
+void * mallocCanFail( size_t size )
+{
+    __CPROVER_assert( size < CBMC_MAX_OBJECT_SIZE, "mallocCanFail size is too big" );
+    return nondet_bool() ? NULL : malloc( size );
```
yourslab, line 35: This used to be how we write these proofs, but it might be better to explicitly say in the harness whether a variable can be `NULL`:
https://github.com/FreeRTOS/coreHTTP/pull/38/files

test/cbmc/proofs/MQTTAgent_Init/MQTTAgent_Init_harness.c
```diff
@@ -0,0 +1,58 @@
...
+  void * pIncomingPacketContext;
+
+  pMqttAgentContext = malloc( sizeof( MQTTAgentContext_t ) );
+  pMsgInterface = mallocCanFail( sizeof( AgentMessageInterface_t ) );
+  pNetworkBuffer = malloc( sizeof( MQTTFixedBuffer_t ) );
+  pTransportInterface = malloc( sizeof( TransportInterface_t ) );
+  /* The incoming packet context can be a pointer to any type.
+   * In this harness, we are using int32_t as that will be good enough
+   * for memory safety proofs. */
+  pIncomingPacketContext = malloc( sizeof( int32_t ) );
```
yourslab, line 49: I don't think we need to initialize `pIncomingPacketContext` at all. Since it can point to a variable of any type and not just an `int32_t`, it should just remain uninitialized so that it has no constraints.

test/cbmc/proofs/MQTTAgent_Init/MQTTAgent_Init_harness.c
```diff
@@ -0,0 +1,58 @@
...
+  MQTTAgentContext_t * pMqttAgentContext;
+  AgentMessageInterface_t * pMsgInterface;
+  MQTTFixedBuffer_t * pNetworkBuffer;
+  TransportInterface_t * pTransportInterface;
+  MQTTGetCurrentTimeFunc_t getCurrentTimeMs;
+  IncomingPublishCallback_t incomingCallback;
+  void * pIncomingPacketContext;
+
+  pMqttAgentContext = malloc( sizeof( MQTTAgentContext_t ) );
+  pMsgInterface = mallocCanFail( sizeof( AgentMessageInterface_t ) );
```
yourslab, line 43: I vaguely remember that CBMC was updated so that `malloc` returns `NULL` by default, so maybe this proof will still work by just using `malloc` instead of `mallocCanFail`. It might be an option that you turn on in the `Makefile` so i'm not quite sure.

test/cbmc/proofs/MQTTAgent_Init/MQTTAgent_Init_harness.c
```diff
@@ -0,0 +1,58 @@
+/*
+ * FreeRTOS V202011.00
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
```
yourslab, line 3: This may not be the right file header. Might also want to change all our file headers to 2021 soon.

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: Thanks. I think I wanted it to be NULL as well. I think as you mentioned in a comment below, malloc just does that. Verified that already. Removing this implementation

test/cbmc/proofs/MQTTAgent_Init/MQTTAgent_Init_harness.c
leegeth: Yea. Good catch. I think that will be needed for all our repos.

yourslab (approved):

aggarg (approved):

PR #6, opened 2021-03-18
[MQTT Agent] : MqttAgent Init and Resume Session Unit Tests

gshvang:


test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    /* Check that only acknowledgements with invalid packet Ids are not eligible for clearance. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+
+    ackInfo.packetId = 1;
+    ackInfo.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that only acknowledgements with valid packet Ids are cleared. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
```
muneebahmed10, line 435: Where are you checking that the acknowledgments are cleared?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+
+    /* Check that MQTTBadParameter is returned if any NULL parameters are passed. */
+    mqttStatus = MQTTAgent_Init( NULL, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, NULL, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, NULL, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
```
muneebahmed10, line 186: Did we skip a NULL networkBuffer?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    msgCtx.recv = receive;
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    msgCtx.releaseCommand = commandRelease;
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+}
```
muneebahmed10, line 214: NULL `incomingCallback`?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -30,25 +30,80 @@
...
 
 /**
- * @brief Time at the beginning of each test. Note that this is not updated with
- * a real clock. Instead, we simply increment this variable.
+ * @brief A mocked send function to send command to agent.
  */
-static uint32_t globalEntryTime = 0;
+static bool( send )( AgentMessageContext_t * pMsgCtx,
+                     const void * pData,
+                     uint32_t blockTimeMs )
```
muneebahmed10, line 59: This is formatted differently from how our functions usually are
```suggestion
static bool send( AgentMessageContext_t * pMsgCtx,
                  const void * pData,
                  uint32_t blockTimeMs )
```
Also, the name `send` is a big generic. `mockSend`?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+
+
+    command.pArgs = &args;
+    mqttAgentContext.mqttContext.nextPacketId = 1;
+    ackInfo.packetId = 1;
+    ackInfo.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+
+    /* Check that publish ack packet id matches the next packet id in the mqttContext. */
```
muneebahmed10, line 353: Where is this checked?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    MQTT_PublishToResend_IgnoreAndReturn( 2 );
```
muneebahmed10, line 354: Why 2?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    command.pArgs = &args;
+    mqttAgentContext.mqttContext.nextPacketId = 1;
+
+    ackInfo.packetId = 0;
+    ackInfo.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+
+    /* Check that publish ack is resent successfully when session resumes. */
+    MQTT_PublishToResend_IgnoreAndReturn( 0 );
+    MQTT_PublishToResend_IgnoreAndReturn( MQTT_PACKET_ID_INVALID );
```
muneebahmed10, line 395: These two lines are identical and so one of them is unnecessary due to the Ignore and return.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    /* Check that publish ack packet id matches the next packet id in the mqttContext. */
+    MQTT_PublishToResend_IgnoreAndReturn( 2 );
+    MQTT_PublishToResend_IgnoreAndReturn( MQTT_PACKET_ID_INVALID );
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, sessionPresent );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+}
+
+
+
+void test_MQTTAgent_ResumeSession_publish_resend_success( void )
```
muneebahmed10, line 362: The code path tested here is identical to `test_MQTTAgent_ResumeSession_session_present_with_packetId_Invalid`

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+
+    /* Check that publish ack is resent successfully when session resumes. */
+    MQTT_PublishToResend_IgnoreAndReturn( 0 );
+    MQTT_PublishToResend_IgnoreAndReturn( MQTT_PACKET_ID_INVALID );
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, sessionPresent );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+}
+
+
+void test_MQTTAgent_ResumeSession_no_session_present_with_broker( void )
```
muneebahmed10, line 401: Name's a bit long
```suggestion
void test_MQTTAgent_ResumeSession_no_session_present( void )
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    msgCtx.releaseCommand = commandRelease;
+    msgCtx.getCommand = getCommand;
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+
+    command.pArgs = &args;
+    mqttAgentContext.mqttContext.nextPacketId = 1;
+
+    ackInfo.packetId = 0;
```
muneebahmed10, line 389: This isn't valid. `MQTT_PACKET_ID_INVALID` == 0

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+
+
+    command.pArgs = &args;
+    mqttAgentContext.mqttContext.nextPacketId = 1;
+    ackInfo.packetId = 1;
+    ackInfo.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that failed resending publish return MQTTBadParameter */
+    MQTT_PublishToResend_IgnoreAndReturn( 1 );
+    MQTT_Publish_IgnoreAndReturn( MQTTBadParameter );
```
muneebahmed10, line 316: I'd prefer `MQTTSendFailed` here since this shouldn't ever return `MQTTBadParameter`

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    TEST_ASSERT_EQUAL( MQTTIllegalState, mqttStatus );
+
+    /* Check that MQTTIllegalState is returned if packet id is 0 */
+    mqttAgentContext.mqttContext.nextPacketId = 0;
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, sessionPresent );
+    TEST_ASSERT_EQUAL( MQTTIllegalState, mqttStatus );
+}
+
+
+void test_MQTTAgent_ResumeSession_session_present_with_packetId_Invalid( void )
```
muneebahmed10, line 254: A better name for this would be `test_MQTTAgent_ResumeSession_no_resent_publishes`, since that's what a 0 return from `MQTT_PublishToResend` indicates. However, I think it would make more sense to combine this function with `test_MQTTAgent_ResumeSession_publish_resend_success` and `test_MQTTAgent_ResumeSession_ack_packetId_not_match` since they all test success cases dealing with various packet IDs.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that only acknowledgements with valid packet Ids are cleared. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+
+    command.pCommandCompleteCallback = CommandCallback;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that command callback is called if it is specified to indicate network error. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
```
muneebahmed10, line 441: We should check that the callback is actually invoked

test/unit-test/mqtt_agent_utest.c
```diff
@@ -30,25 +30,80 @@
...
+{
+    return;
+}
+
+/**
+ * @brief Context Definition with which tasks may deliver messages to the agent.
+ */
+struct AgentMessageContext
+{
+    int mock;
```
muneebahmed10, line 105: I'd suggest changing this to something that would be more useful when testing `MQTTAgent_CommandLoop`
```suggestion
    Command_t * pSentCommand;
```
Then `send` can set this pointer and `receive` can read from it

test/unit-test/mqtt_agent_utest.c
```diff
@@ -30,25 +30,80 @@
...
+ */
+static bool( receive )( AgentMessageContext_t * pMsgCtx,
+                        void * pBuffer,
+                        uint32_t blockTimeMs )
+{
+    return true;
+}
+
+/**
+ * @brief A mocked function to obtain a pointer to an allocated command.
```
muneebahmed10, line 75: This description is incorrect

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    msgCtx.pMsgCtx = &msg;
+    msgCtx.send = send;
+    msgCtx.recv = receive;
+    msgCtx.releaseCommand = commandRelease;
+    msgCtx.getCommand = getCommand;
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+
+
+    /* Check that MQTTIllegalState is returned if any NULL mqttAgentContext is passed. */
```
muneebahmed10, line 243: This really should be `MQTTBadParameter`. We should fix this in the code

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
 
 void test_MQTTAgent_Init_Happy_Path( void )
 {
+    MQTTAgentContext_t mqttAgentContext;
+    AgentMessageInterface_t msgCtx;
+    MQTTFixedBuffer_t networkBuffer;
+    TransportInterface_t transportInterface;
+    IncomingPublishCallback_t incomingCallback;
```
muneebahmed10, line 143: We should initialize this since it's a function pointer. Actually, we should probably initialize everything, even if it's just to 0.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+    msgCtx.pMsgCtx = &msg;
+    msgCtx.send = send;
+    msgCtx.recv = receive;
+    msgCtx.releaseCommand = commandRelease;
+    msgCtx.getCommand = getCommand;
 
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL_PTR( incomingCallback, mqttAgentContext.pIncomingCallback );
```
muneebahmed10, line 159: This should actually not return success if this is NULL. We should fix the code for this

test/unit-test/mqtt_agent_utest.c
gshvang: Yeah

test/unit-test/mqtt_agent_utest.c
```diff
@@ -30,25 +30,80 @@
...
+    return true;
+}
+
+/**
+ * @brief A mocked function to release an allocated command.
+ */
+static Command_t * ( getCommand ) ( uint32_t blockTimeMs )
+{
+    Command_t * command;
+    return command;
```
yourslab, line 88: `command` is not initialized here, so this might lead to returning a `NULL` pointer sometimes. I think it would be better to have a static declaration.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +136,307 @@ int suiteTearDown( int numFailures )
...
+ */
+void test_MQTTAgent_ResumeSession_Invalid_Params( void )
+{
+    MQTTAgentContext_t mqttAgentContext;
+    bool sessionPresent = true;
+    MQTTStatus_t mqttStatus;
+
+    AgentMessageInterface_t msgCtx;
+    MQTTFixedBuffer_t networkBuffer;
+    TransportInterface_t transportInterface;
```
yourslab, line 227: We never initialize the `transportInterface`, so do we ever check if its members are `NULL`?

yourslab (commented):
It's part of our coding standard to initialize variables to avoid non-deterministic execution paths.

test/unit-test/mqtt_agent_utest.c
muneebahmed10: We should zero initialize it, but its members don't matter since the agent doesn't use it. It's only passed to MQTT_Init(), which is mocked

test/unit-test/mqtt_agent_utest.c
muneebahmed10: I take this back. MQTT_Init() will check the network buffer for us, so we shouldn't need to check it twice. A comment here about why we're not testing the network buffer will be sufficient.

test/unit-test/mqtt_agent_utest.c
gshvang: I have initialized it here.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +139,276 @@ int suiteTearDown( int numFailures )
...
+    ackInfo.packetId = MQTT_PACKET_ID_INVALID;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that acknowledgements with invalid packet Ids are not eligible for clearance. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL_MEMORY( &ackInfo, &mqttAgentContext.pPendingAcks[ 0 ], sizeof( AckInfo_t ) );
+
+    ackInfo_1.packetId = 1;
+    ackInfo_1.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 1 ] = ackInfo_1;
```
muneebahmed10, line 399: Do we even need `ackInfo_1`?
```suggestion
    mqttAgentContext.pPendingAcks[ 1 ].packetId = 1U;
    mqttAgentContext.pPendingAcks[ 1 ].pOriginalCommand = &command;
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +139,276 @@ int suiteTearDown( int numFailures )
...
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+
+    mqttAgentContext.mqttContext.nextPacketId = 1;
+    ackInfo.packetId = MQTT_PACKET_ID_INVALID;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+    /* Check that acknowledgements with invalid packet Ids are not eligible for clearance. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL_MEMORY( &ackInfo, &mqttAgentContext.pPendingAcks[ 0 ], sizeof( AckInfo_t ) );
```
muneebahmed10, line 395: This was already true before the call, so what are we testing exactly? Entries with invalid packet IDs are empty

test/unit-test/mqtt_agent_utest.c
yourslab: This was from earlier

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +139,276 @@ int suiteTearDown( int numFailures )
...
+    void * incomingPacketContext;
+    AgentMessageContext_t msg;
+
+    msgCtx.pMsgCtx = &msg;
+    msgCtx.send = mockSend;
+    msgCtx.recv = mockReceive;
+    msgCtx.releaseCommand = mockCommandRelease;
+    msgCtx.getCommand = mockGetCommand;
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgCtx, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
```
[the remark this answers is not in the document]
gshvang: Its good, I will incorporate this.

test/unit-test/mqtt_agent_utest.c
gshvang: This test case was not covering any branch so removed it.

test/unit-test/mqtt_agent_utest.c
gshvang: Resolved with static declaration.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,196 @@ int suiteTearDown( int numFailures )
...
+    command.pArgs = &args;
+
+    ackInfo.packetId = 1U;
+    ackInfo.pOriginalCommand = &command;
+    mqttAgentContext.pPendingAcks[ 0 ] = ackInfo;
+
+    /* Check that publish ack is resent successfully when session resumes. */
+    MQTT_PublishToResend_IgnoreAndReturn( 1 );
+    MQTT_Publish_IgnoreAndReturn( MQTTSuccess );
+    MQTT_PublishToResend_IgnoreAndReturn( MQTT_PACKET_ID_INVALID );
```
muneebahmed10, line 355: Ignore and return overrides the previous use of ignore and return, no? Shouldn't this be "ExpectAnyArgsAndReturn"?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,196 @@ int suiteTearDown( int numFailures )
...
+    msgInterface.pMsgCtx = &msg;
+    msgInterface.send = mockSend;
+    msgInterface.recv = mockReceive;
+    msgInterface.releaseCommand = mockReleaseCommand;
+    msgInterface.getCommand = mockGetCommand;
 
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTSuccess );
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgInterface, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL_PTR( incomingCallback, mqttAgentContext.pIncomingCallback );
```
muneebahmed10, line 213: What is the value of `incomingCallback` here?

source/mqtt_agent.c
```diff
@@ -809,15 +809,18 @@ MQTTStatus_t MQTTAgent_Init( MQTTAgentContext_t * pMqttAgentContext,
...
         returnStatus = MQTTBadParameter;
     }
     else if( ( pMsgInterface->pMsgCtx == NULL ) ||
              ( pMsgInterface->send == NULL ) ||
              ( pMsgInterface->recv == NULL ) ||
-             ( pMsgInterface->getCommand == NULL ) ||
-             ( pMsgInterface->releaseCommand == NULL ) )
+             ( pMsgInterface->releaseCommand == NULL ) ||
+             ( pMsgInterface->getCommand == NULL ) )
+
```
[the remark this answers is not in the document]
yourslab: I think we have reintroduced the extra newline here.

source/mqtt_agent.c
```diff
@@ -809,15 +809,18 @@ MQTTStatus_t MQTTAgent_Init( MQTTAgentContext_t * pMqttAgentContext,
     if( ( pMqttAgentContext == NULL ) ||
         ( pMsgInterface == NULL ) ||
         ( pTransportInterface == NULL ) ||
-        ( getCurrentTimeMs == NULL ) )
+        ( getCurrentTimeMs == NULL ) ||
+        ( pNetworkBuffer == NULL ) ||
```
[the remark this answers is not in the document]
muneebahmed10: We should remove the check for the network buffer

test/unit-test/mqtt_agent_utest.c
muneebahmed10: Unused variable? It looks like we can delete this

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,197 @@ int suiteTearDown( int numFailures )
...
+    mqttStatus = MQTTAgent_Init( NULL, &msgInterface, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, NULL, &networkBuffer, &transportInterface, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    mqttStatus = MQTTAgent_Init( &mqttAgentContext, &msgInterface, &networkBuffer, NULL, getTime, incomingCallback, incomingPacketContext );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    MQTT_Init_ExpectAnyArgsAndReturn( MQTTBadParameter );
```
muneebahmed10: A comment would be helpful
```suggestion
    /* MQTT_Init() should check the network buffer. */
    MQTT_Init_ExpectAnyArgsAndReturn( MQTTBadParameter );
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,197 @@ int suiteTearDown( int numFailures )
...
  * @brief Test that any NULL parameter causes MQTTAgent_Init to return MQTTBadParameter.
  */
-void test_MQTT_Init_Invalid_Params( void )
+void test_MQTTAgent_Init_Invalid_Params( void )
+{
+    MQTTAgentContext_t mqttAgentContext;
+    AgentMessageInterface_t msgInterface = { 0 };
+    MQTTFixedBuffer_t networkBuffer = { 0 };
+    TransportInterface_t transportInterface = { 0 };
+    IncomingPublishCallback_t incomingCallback;
```
muneebahmed10, line 227: This should be initialized
```suggestion
    IncomingPublishCallback_t incomingCallback = mockPublishCallback;
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,200 @@ int suiteTearDown( int numFailures )
...
+    mqttAgentContext.pPendingAcks[ 1 ].packetId = 1U;
+    mqttAgentContext.pPendingAcks[ 1 ].pOriginalCommand = &command;
+    /* Check that only acknowledgments with valid packet IDs are cleared. */
+    mqttStatus = MQTTAgent_ResumeSession( &mqttAgentContext, false );
+    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL( 0, mqttAgentContext.pPendingAcks[ 1 ].packetId );
+    TEST_ASSERT_EQUAL( NULL, mqttAgentContext.pPendingAcks[ 1 ].pOriginalCommand );
+
+    command.pCommandCompleteCallback = mockCommandCallback;
+    ackInfoCallback.pOriginalCommand = &command;
```
muneebahmed10, line 384: What is the use of this variable?

gshvang: It was left by mistake. I removed it

muneebahmed10:
Looking at the coverage report, it looks like coverage is missing for when the following conditions are evaluated as false:
https://github.com/FreeRTOS/coreMQTT-Agent/blob/1d6c07a77e8ad98e08417dac573ff1cbd30fb30d/source/mqtt_agent.c#L744
https://github.com/FreeRTOS/coreMQTT-Agent/blob/1d6c07a77e8ad98e08417dac573ff1cbd30fb30d/source/mqtt_agent.c#L783

If we're not adding them in this PR, we should make a note of it, though it might make sense to add them now since MQTTAgent_ResumeSession() is the only function to use those functions

muneebahmed10 (approved):
Approving, but I'd prefer if we get 100% branch coverage for the 2 cases noted above

test/unit-test/mqtt_agent_utest.c
```diff
@@ -81,14 +192,199 @@ int suiteTearDown( int numFailures )
...
+}
+
+
+void test_MQTTAgent_ResumeSession_no_session_present( void )
+{
+    bool sessionPresent = true;
+    MQTTStatus_t mqttStatus;
+    MQTTAgentContext_t mqttAgentContext;
+    Command_t command;
+    AckInfo_t ackInfo, ackInfoCallback;
```
muneebahmed10, line 371: These are unused
```suggestion
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -313,6 +313,20 @@ void test_MQTTAgent_ResumeSession_session_present_no_resent_publishes( void )
...
+void test_MQTTAgent_ResumeSession_session_present_no_puback_found( void )
+{
+    bool sessionPresent = true;
+    MQTTStatus_t mqttStatus;
+    MQTTAgentContext_t mqttAgentContext;
+
+    setupAgentContext( &mqttAgentContext );
+
+    MQTT_PublishToResend_ExpectAnyArgsAndReturn( 2 );
+    mqttAgentContext.pPendingAcks[ 0 ].packetId=1U;
```
muneebahmed10, line 325: Uncrustify
```suggestion
    mqttAgentContext.pPendingAcks[ 0 ].packetId = 1U;
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -313,6 +313,20 @@ void test_MQTTAgent_ResumeSession_session_present_no_resent_publishes( void )
     TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
 }
 
+void test_MQTTAgent_ResumeSession_session_present_no_puback_found( void )
```
muneebahmed10: I'd prefer something like
```suggestion
void test_MQTTAgent_ResumeSession_session_present_no_publish_found( void )
```
since it's the record of the publish we're storing, not the puback

gshvang:
> Approving, but I'd prefer if we get 100% branch coverage for the 2 cases noted above

Included the coverage for he 2 cases above.

test/unit-test/mqtt_agent_utest.c
gshvang: Resolved

muneebahmed10 (approved):

yourslab (approved):

n9wxu (approved):

PR #7, opened 2021-03-18
Add input validation and some unit tests

muneebahmed10:
*Description*:
Adds input validation to most MQTT Agent API functions, and unit tests for the following:
* `MQTTAgent_Subscribe`
* `MQTTAgent_Unsubscribe`
* `MQTTAgent_ProcessLoop`
* `MQTTAgent_Disconnect`
* `MQTTAgent_Ping`
* `MQTTAgent_Terminate`

In the process of writing tests, some changes were made to the source code:
* Some more `NULL` checks.
* `assert`s in favor of `if` checks where the parameter had already been checked for `NULL`
* Some missed calls in `MQTTAgentCommand_Connect` to resume sessions.

`MQTTAgent_Connect` and `MQTTAgent_Publish` still need unit tests, however I thought those should be added later lest this PR grow too large. Coverage of `mqtt_agent.c` is around 33%.

### DIFF
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -408,7 +408,7 @@ static MQTTStatus_t createCommand( CommandType_t commandType,
             }
 
             /* Will the message fit in the defined buffer? */
-            isValid = ( ( pPublishInfo->payloadLength + uxHeaderBytes ) < pMqttAgentContext->mqttContext.networkBuffer.size ) &&
+            isValid = ( uxHeaderBytes < pMqttAgentContext->mqttContext.networkBuffer.size ) &&
                       ( isSpace == true );
 
             break;
@@ -807,7 +807,6 @@ static bool validateStruct( MQTTAgentContext_t * pMqttAgentContext,
         LogError( ( "Pointer cannot be NULL. pMqttAgentContext=%p, pCommandInfo=%p.",
                     ( void * ) pMqttAgentContext,
                     ( void * ) pCommandInfo ) );
-        ret = false;
     }
     else if( ( pMqttAgentContext->agentInterface.send == NULL ) ||
              ( pMqttAgentContext->agentInterface.recv == NULL ) ||
@@ -816,7 +815,6 @@ static bool validateStruct( MQTTAgentContext_t * pMqttAgentContext,
              ( pMqttAgentContext->agentInterface.pMsgCtx == NULL ) )
     {
         LogError( ( "pMqttAgentContext must have initialized its messaging interface." ) );
-        ret = false;
     }
     else
     {
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -218,6 +218,30 @@ static MQTTStatus_t resendPublishes( MQTTAgentContext_t * pMqttAgentContext );
 static void clearPendingAcknowledgments( MQTTAgentContext_t * pMqttAgentContext );
 
 /**
+ * @brief Validate an #MQTTAgentContext_t and a #CommandInfo_t from API
+ * functions.
+ *
+ * @param[in] pMqttAgentContext #MQTTAgentContext_t to validate.
+ * @param[in] pCommandInfo #CommandInfo_t to validate.
+ *
+ * @return `true` if parameters are valid, else `false`.
+ */
+static bool validateStruct( MQTTAgentContext_t * pMqttAgentContext,
+                            CommandInfo_t * pCommandInfo );
+
+/**
+ * @brief Validate the parameters for a CONNECT, SUBSCRIBE, UNSUBSCRIBE
+ * or PUBLISH.
+ *
+ * @param[in] commandType CONNECT, SUBSCRIBE, UNSUBSCRIBE, or PUBLISH.
+ * @param[in] pParams Parameter structure to validate.
+ *
+ * @return `true` if parameter structure is valid, else `false`.
+ */
+static bool validateParams( CommandType_t commandType,
+                            void * pParams );
+
+/**
  * @brief Called before accepting any PUBLISH or SUBSCRIBE messages to check
  * there is space in the pending ACK list for the outgoing PUBLISH or SUBSCRIBE.
  *
@@ -244,20 +268,19 @@ static bool isSpaceInPendingAckList( MQTTAgentContext_t * pAgentContext )
     bool spaceFound = false;
     size_t i;
 
-    if( pAgentContext != NULL )
-    {
-        pendingAcks = pAgentContext->pPendingAcks;
+    assert( pAgentContext != NULL );
 
-        /* Are there any open slots? */
-        for( i = 0; i < MQTT_AGENT_MAX_OUTSTANDING_ACKS; i++ )
+    pendingAcks = pAgentContext->pPendingAcks;
+
+    /* Are there any open slots? */
+    for( i = 0; i < MQTT_AGENT_MAX_OUTSTANDING_ACKS; i++ )
+    {
+        /* If the packetId is MQTT_PACKET_ID_INVALID then the array space is
+         * not in use. */
+        if( pendingAcks[ i ].packetId == MQTT_PACKET_ID_INVALID )
         {
-            /* If the packetId is MQTT_PACKET_ID_INVALID then the array space is
-             * not in use. */
-            if( pendingAcks[ i ].packetId == MQTT_PACKET_ID_INVALID )
-            {
-                spaceFound = true;
-                break;
-            }
+            spaceFound = true;
+            break;
         }
     }
 
@@ -344,27 +367,17 @@ static MQTTStatus_t createCommand( CommandType_t commandType,
     size_t uxHeaderBytes;
     const size_t uxControlAndLengthBytes = ( size_t ) 4; /* Control, remaining length and length bytes. */
 
+    assert( pMqttAgentContext != NULL );
+    assert( pCommand != NULL );
+
     memset( pCommand, 0x00, sizeof( Command_t ) );
 
     /* Determine if required parameters are present in context. */
     switch( commandType )
     {
         case SUBSCRIBE:
-
-            /* This message type results in the broker returning an ACK.  The
-             * agent maintains an array of outstanding ACK messages.  See if
-             * the array contains space for another outstanding ack. */
-            isSpace = isSpaceInPendingAckList( pMqttAgentContext );
-
-            pSubscribeArgs = ( MQTTAgentSubscribeArgs_t * ) pMqttInfoParam;
-            isValid = ( pSubscribeArgs != NULL ) &&
-                      ( pSubscribeArgs->pSubscribeInfo != NULL ) &&
-                      ( pMqttAgentContext != NULL ) &&
-                      ( isSpace == true );
-
-            break;
-
         case UNSUBSCRIBE:
+            assert( pMqttInfoParam != NULL );
 
             /* This message type results in the broker returning an ACK.  The
              * agent maintains an array of outstanding ACK messages.  See if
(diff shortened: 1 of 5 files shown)


source/mqtt_agent.c
```diff
@@ -344,81 +367,58 @@ static MQTTStatus_t createCommand( CommandType_t commandType,
...
-                    isSpace = isSpaceInPendingAckList( pMqttAgentContext );
-                }
 
-                isValid = ( pMqttAgentContext != NULL ) &&
-                          /* Will the message fit in the defined buffer? */
-                          ( ( pPublishInfo->payloadLength + uxHeaderBytes ) < pMqttAgentContext->mqttContext.networkBuffer.size ) &&
-                          ( isSpace == true );
-            }
+            /* Will the message fit in the defined buffer? */
+            isValid = ( ( pPublishInfo->payloadLength + uxHeaderBytes ) < pMqttAgentContext->mqttContext.networkBuffer.size ) &&
```
leegeth, line 411: I think the payload will not be copied over to the MQTT network buffer for publish. Do we need to consider the payload length in that case?

source/mqtt_agent.c
```diff
@@ -796,6 +796,76 @@ static void clearPendingAcknowledgments( MQTTAgentContext_t * pMqttAgentContext
...
+{
+    bool ret = false;
+
+    if( ( pMqttAgentContext == NULL ) ||
+        ( pCommandInfo == NULL ) )
+    {
+        LogError( ( "Pointer cannot be NULL. pMqttAgentContext=%p, pCommandInfo=%p.",
+                    ( void * ) pMqttAgentContext,
+                    ( void * ) pCommandInfo ) );
+        ret = false;
```
leegeth, line 810: This line can be removed.

source/mqtt_agent.c
```diff
@@ -796,6 +796,76 @@ static void clearPendingAcknowledgments( MQTTAgentContext_t * pMqttAgentContext
...
+    }
+    else if( ( pMqttAgentContext->agentInterface.send == NULL ) ||
+             ( pMqttAgentContext->agentInterface.recv == NULL ) ||
+             ( pMqttAgentContext->agentInterface.getCommand == NULL ) ||
+             ( pMqttAgentContext->agentInterface.releaseCommand == NULL ) ||
+             ( pMqttAgentContext->agentInterface.pMsgCtx == NULL ) )
+    {
+        LogError( ( "pMqttAgentContext must have initialized its messaging interface." ) );
+        ret = false;
```
leegeth, line 819: same here

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,449 @@ int suiteTearDown( int numFailures )
...
+    agentContext.agentInterface.releaseCommand = mockReleaseCommand;
+    agentContext.agentInterface.pMsgCtx = NULL;
+    mqttStatus = MQTTAgent_Terminate( &agentContext, &commandInfo );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
+
+    agentContext.agentInterface.pMsgCtx = &globalMessageContext;
+    /* Invalid packet ID to indicate uninitialized context. */
+    agentContext.mqttContext.nextPacketId = MQTT_PACKET_ID_INVALID;
+    mqttStatus = MQTTAgent_Terminate( &agentContext, &commandInfo );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
```
leegeth, line 303: Why is this testing limited to the MQTTAgent_Terminate API for create command testing? Is it testing all the paths for bad parameters for create command?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,449 @@ int suiteTearDown( int numFailures )
...
+
+    /* Use Ping since it's one that doesn't have an additional pointer param. */
+    pCommandToReturn = NULL;
+    mqttStatus = MQTTAgent_Ping( &agentContext, &commandInfo );
+    TEST_ASSERT_EQUAL( MQTTNoMemory, mqttStatus );
+
+    pCommandToReturn = &command;
+    agentContext.agentInterface.send = mockSendFail;
+    mqttStatus = MQTTAgent_Ping( &agentContext, &commandInfo );
+    TEST_ASSERT_EQUAL( MQTTSendFailed, mqttStatus );
```
leegeth, line 327: I think it is better to either place the send failed in a different test or rename the test as the test name indicates allocation error.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,449 @@ int suiteTearDown( int numFailures )
...
+{
+    MQTTAgentContext_t agentContext;
+    MQTTStatus_t mqttStatus;
+    CommandInfo_t commandInfo = { 0 };
+    Command_t command = { 0 };
+
+    setupAgentContext( &agentContext );
+    pCommandToReturn = &command;
+
+    /* We have already tested generic cases of NULL parameters above,
```
leegeth, line 466: please be specific about the test in which we are doing this verification as that will be helpful for a future reader.

test/unit-test/mqtt_agent_utest.c
muneebahmed10: Every single thread-safe API function is a wrapper around the same static function to create commands and add to the queue. I thought it redundant to test these cases for all 8 of them, so I just tested this one exhaustively and the others for coverage. Terminate just happened to be the one I picked

source/mqtt_agent.c
muneebahmed10: Yes that's true, so I guess not

source/mqtt_agent.c
muneebahmed10: That's true, but I didn't know how MISRA would react since the block would be empty otherwise (since logs are compiled out). I removed it for now and can add it back if it causes an issue

test/unit-test/mqtt_agent_utest.c
muneebahmed10: Renamed to "creation error"

leegeth (approved):

source/mqtt_agent.c
```diff
@@ -217,6 +217,30 @@ static MQTTStatus_t resendPublishes( MQTTAgentContext_t * pMqttAgentContext );
  */
```
alfred2g, line 217: it would be nicer if you split the tests from the code
1- the commit title would be more relevant - easier to find commits and what they actually did
2- if you had to revert the impact would be smaller and re-merging easier
3- single responsibility of commits
..

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,455 @@ int suiteTearDown( int numFailures )
...
+    ( void ) pMsgCtx;
+    ( void ) pData;
+    ( void ) blockTimeMs;
+    return false;
+}
 
+/**
+ * @brief A mocked receive function for the agent to receive commands.
+ */
+static bool mockReceive( AgentMessageContext_t * pMsgCtx,
```
alfred2g, line 145: if a function returns only one thing, it should be called stub not mock

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,455 @@ int suiteTearDown( int numFailures )
...
+    Command_t ** pCommandToReceive = ( Command_t ** ) pBuffer;
 
+    *pCommandToReceive = pMsgCtx->pSentCommand;
+    return true;
+}
+
+/**
+ * @brief A mocked function to obtain an allocated command.
+ */
+static Command_t * mockGetCommand( uint32_t blockTimeMs )
```
alfred2g, line 158: same as above.. stub

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,455 @@ int suiteTearDown( int numFailures )
...
+{
+    return pCommandToReturn;
+}
+
+/**
+ * @brief A mocked function to release an allocated command.
+ */
+static bool mockReleaseCommand( Command_t * pCommandToRelease )
```
alfred2g, line 166: same

alfred2g: also for the rest of them
some arguably might be called fake since they do some processing ( if, ++ ), maybe not enough

test/unit-test/mqtt_agent_utest.c
aggarg: This seems like we are using the knowledge of the implementation to decide which API to test to achieve coverage numbers . Tests should be written against APIs - what if I update one `MQTTAgent_` API and stop calling the common function which checks for error. Would it introduced an error without failing a test.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,455 @@ int suiteTearDown( int numFailures )
...
+    subscribeArgs.pSubscribeInfo = &subscribeInfo;
+    subscribeArgs.numSubscriptions = 1U;
+    mqttStatus = MQTTAgent_Subscribe( &agentContext, &subscribeArgs, &commandInfo );
+    TEST_ASSERT_EQUAL( MQTTNoMemory, mqttStatus );
+}
+
+/**
+ * @brief Test that MQTTAgent_Subscribe() works as intended.
+ */
+void test_MQTTAgent_Subscribe( void )
```
aggarg, line 367: Can be broken in three separate tests - NULL params, incorrect params, happy path.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,455 @@ int suiteTearDown( int numFailures )
...
+    TEST_ASSERT_EQUAL_PTR( &command, globalMessageContext.pSentCommand );
+    TEST_ASSERT_EQUAL( SUBSCRIBE, command.commandType );
+    TEST_ASSERT_EQUAL_PTR( &subscribeArgs, command.pArgs );
+    TEST_ASSERT_NULL( command.pCommandCompleteCallback );
+}
+
+/**
+ * @brief Test that MQTTAgent_Unsubscribe() works as intended.
+ */
+void test_MQTTAgent_Unsubscribe( void )
```
aggarg, line 411: Same as above.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,536 @@ int suiteTearDown( int numFailures )
...
-    
+}
+
+/* ========================================================================== */
+
+/**
+ * @brief Test error cases when a command cannot be obtained or sent
+ * for API functions that create commands.
+ */
+void test_MQTTAgent_API_Create_Command_Creation_Error( void )
```
aggarg, line 315: The name is misleading. This test should be broken into two:

1. MQTTAgent_Ping_Queue_Full [or whatever the reason for out of memory is].
2. MQTTAgent_Ping_Send_To_Queue_Failed 

The idea is that a test should test only one thing and it  must only fail when that one thing is false.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -74,21 +110,536 @@ int suiteTearDown( int numFailures )
...
+    /* Also test that the command was set. */
+    TEST_ASSERT_EQUAL( PING, command.commandType );
+}
+
+/**
+ * @brief Test that a bad parameter is returned when there
+ * is no more space to store a pending acknowledgment for
+ * a command that expects one.
+ */
+void test_MQTTAgent_API_Create_Command_No_Ack_Space( void )
```
aggarg, line 344: The name should be MQTTAgent_Subscribe_No_Ack_Space

aggarg (approved):

aggarg (approved):

PR #8, opened 2021-03-19
CBMC proof for MQTTAgent_Terminate API

leegeth:
CBMC proof for MQTTAgent_Terminate API.


yourslab (approved):

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
```diff
@@ -0,0 +1,39 @@
...
+
+void harness()
+{
+    MQTTAgentContext_t * pMqttAgentContext;
+    CommandInfo_t * pCommandInfo;
+
+    pMqttAgentContext = allocateMqttAgentContext( NULL );
+    __CPROVER_assume( isValidMqttAgentContext( pMqttAgentContext ) );
+
+    pCommandInfo = malloc( sizeof( CommandInfo_t ) );
```
yourslab, line 35: Same comment from PR #9. I think we will need a stub for `cmdCompleteCallback`.

leegeth: MQTT_Terminate is a function which just attempts to push a command to the queue. So a non deterministic value will be fine for command callback for the proof

yourslab (approved):

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
```diff
@@ -0,0 +1,39 @@
...
+
+    MQTTAgent_Terminate( pMqttAgentContext,
```
dan4thewin, line 37: Is there anything we can assert after the function returns?

test/cbmc/proofs/MQTTAgent_Terminate/README.md
```diff
@@ -0,0 +1,20 @@
+MQTTAgent_Terminate proof
+==============
+
+This directory contains a memory safety proof for MQTTAgent_Terminate.
```
dan4thewin, line 4: Can we say anything about the coverage?  Approximate time to run?  What is tested vs. what is stubbed?

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
yourslab: After giving it second thought, it might also be a good idea to explicitly define a callback that returns a `nondet_bool()` so that it is also self-documenting.

test/cbmc/proofs/MQTTAgent_Terminate/README.md
leegeth: Sure. That will be a good addition. Thanks Dan

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
leegeth: I am still of the opinion that it doesn't add much value to this proof as all the members in the CommanInfo_t in this proof are using only a non-deterministic value. Please let me know if you think otherwise.

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
leegeth: I don't think an assert on the function return adds value in this proof as the memory safety will be proven for the implementation of the function with this harness. Please let me know if you think otherwise.

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
leegeth: As discussed offline, added more comments about using non deterministic values for the CommandInfo_t

test/cbmc/proofs/MQTTAgent_Terminate/README.md
leegeth: I have added a statement about the stubs used. I think the time to run will be depending on many factors such as system on which it is running and will be viewable from the results in the CI run. Coverage also will be changing depending on if there are any changes made to the API later. It will also be seen from the CI run. So those are not added in README, but added a mention about the stubs. Please let me know if you agree with this change.

yourslab (approved):

test/cbmc/proofs/MQTTAgent_Terminate/MQTTAgent_Terminate_harness.c
leegeth: As discussed offline, I have added the asserts for the return.

test/cbmc/proofs/MQTTAgent_Terminate/README.md
leegeth: As discussed offline, added comments for these.

PR #10, opened 2021-03-19
CBMC proof for MQTTAgent_Disconnect

leegeth:
CBMC proof for MQTTAgent_Disconnect.

The errors are fixed in the PR #7. This proof should be merged only after the PR #7 is merged and the branch is rebased.

### DIFF
--- a/test/cbmc/sources/mqtt_agent_cbmc_state.c
+++ b/test/cbmc/sources/mqtt_agent_cbmc_state.c
@@ -1,0 +1,139 @@
+/*
+ * coreMQTT-Agent v1.0.0
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file mqtt_agent_cbmc_state.c
+ * @brief Implements the functions defined in mqtt_agent_cbmc_state.h.
+ */
+#include <stdint.h>
+#include <stdlib.h>
+#include "mqtt_agent.h"
+#include "mqtt_agent_cbmc_state.h"
+#include "network_interface_stubs.h"
+#include "get_time_stub.h"
+#include "incoming_publish_callback_stub.h"
+#include "agent_message_stubs.h"
+#include "agent_command_pool_stubs.h"
+
+
+MQTTFixedBuffer_t * allocateMqttFixedBuffer( MQTTFixedBuffer_t * pFixedBuffer )
+{
+    if( pFixedBuffer == NULL )
+    {
+        pFixedBuffer = malloc( sizeof( MQTTFixedBuffer_t ) );
+    }
+
+    if( pFixedBuffer != NULL )
+    {
+        __CPROVER_assume( pFixedBuffer->size < CBMC_MAX_OBJECT_SIZE );
+        pFixedBuffer->pBuffer = malloc( pFixedBuffer->size );
+    }
+
+    return pFixedBuffer;
+}
+
+bool isValidMqttFixedBuffer( const MQTTFixedBuffer_t * pFixedBuffer )
+{
+    bool isValid = true;
+
+    if( pFixedBuffer != NULL )
+    {
+        isValid = pFixedBuffer->size < CBMC_MAX_OBJECT_SIZE;
+    }
+
+    return isValid;
+}
+
+MQTTAgentContext_t * allocateMqttAgentContext( MQTTAgentContext_t * pContext )
+{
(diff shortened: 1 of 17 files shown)


yourslab (approved):

test/cbmc/sources/mqtt_agent_cbmc_state.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+bool isValidMqttAgentContext( const MQTTAgentContext_t * pContext )
+{
+    bool isValid = false;
+
+    if( pContext != NULL )
+    {
+        isValid = true;
+        isValid = isValid && pContext->mqttContext.networkBuffer.size < CBMC_MAX_OBJECT_SIZE;
```
yourslab, line 134: Can this predicate be moved into `isValidMqttFixedBuffer`?

test/cbmc/stubs/network_interface_stubs.c
```diff
@@ -0,0 +1,109 @@
...
+                                     void * pBuffer,
+                                     size_t bytesToRecv )
+{
+    __CPROVER_assert( pBuffer != NULL,
+                      "NetworkInterfaceReceiveStub pBuffer is not NULL." );
+
+    __CPROVER_assert( __CPROVER_w_ok( pBuffer, bytesToRecv ),
+                      "NetworkInterfaceReceiveStub pBuffer is writable up to bytesToRecv." );
+
+    __CPROVER_havoc_object( pBuffer );
```
yourslab, line 55: I'm curious to know more about the purpose of this call. I tried to Google the function but couldn't find much documentation on it.

test/cbmc/sources/mqtt_agent_cbmc_state.c
yourslab: Actually, it seems the check is already there.

test/cbmc/sources/mqtt_agent_cbmc_state.c
```diff
@@ -0,0 +1,139 @@
...
+
+    pNetworkBuffer = allocateMqttFixedBuffer( NULL );
+    __CPROVER_assume( isValidMqttFixedBuffer( pNetworkBuffer ) );
+
+    pMessageInterface = malloc( sizeof( AgentMessageInterface_t ) );
+
+    if( pMessageInterface != NULL )
+    {
+        pMessageInterface->send = AgentMessageSendStub;
+        pMessageInterface->recv = AgentMessageRecvStub;
```
yourslab, line 98: I think these stubs are missing their definition

test/cbmc/sources/mqtt_agent_cbmc_state.c
```diff
@@ -0,0 +1,139 @@
...
+        pMessageInterface->getCommand = AgentGetCommandStub;
+        pMessageInterface->releaseCommand = Agent_ReleaseCommand;
+    }
```
yourslab, line 101: This is fine, but I remember Mark preferred having an `isValidMessageInterface` where we say `__CPROVER_assume(pMessageInterface->releaseCommand == Agent_ReleaseCommand)`. Same for other assignments inside this block.

test/cbmc/sources/mqtt_agent_cbmc_state.c
```diff
@@ -0,0 +1,139 @@
...
+    pTransportInterface = malloc( sizeof( TransportInterface_t ) );
+
+    if( pTransportInterface != NULL )
+    {
+        /* The possibility that recv and send callbacks are NULL is tested in the
+         * MQTT_Init proof. MQTT_Init is required to be called before any other
+         * function in core_mqtt.h. */
+        pTransportInterface->recv = NetworkInterfaceReceiveStub;
+        pTransportInterface->send = NetworkInterfaceSendStub;
+    }
```
yourslab, line 88: Same comment as earlier. It would be great to have an `isValidTransportInterface` where we call `__CPROVER_assume`. However, if you prefer assignment or it runs faster, we can use `allocateTransportInterface` instead.

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: Yea. Good catch. This logic was copied over from the coreMQTT stub implementations. I think this line can be removed as we have a check in the isValidMqttFixedBuffer function

test/cbmc/stubs/network_interface_stubs.c
leegeth: I also searched for it and found only this doc http://cprover.diffblue.com/cprover__builtin__headers_8h.html. The logic for this was from the coreMQTT stub implementation. As per my understanding of the code flow and from the name of the function, the buffer will be filled with some random data which will simulate a read. So I think this is a good addition.

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: Yea that was an intentional choice. I just kept the declaration of the function so that the return will always be an no deterministic value from either true or false( return type of the function). It will be just enough for the proofs.

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: Yea. This is another way to handle it. However from a speed of the proof I prefer the assignment here. Do you know of any benefits of using __CPROVER_assume instead?

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: same response as above.

test/cbmc/sources/mqtt_agent_cbmc_state.c
yourslab: It might still be better to define a stub that returns `nondet_bool()` so that it is self-documenting.

leegeth: Agree. Makes sense. I will add that

test/cbmc/sources/mqtt_agent_cbmc_state.c
yourslab: That would be a good question to ask in the CBMC channel.

yourslab (approved):

test/cbmc/sources/mqtt_agent_cbmc_state.c
leegeth: Yea checking with Mark from ARG team. now. As discussed offline, if he points out about a benefit in that approach, I can create a follow up PR to address that. Thanks

markrtuttle: Generally speaking, the thing I care about most is the set of assumptions that the proof depends on.  I want to know that all assumptions are listed in one place (or in a well-defined and documented set of places).  So I want the proof harness to list the collection of proof assumptions as a list of assumptions.  You can, of course, model assumptions like initializations in the form of assignments in the harness, but that makes the assumptions implicit and harder to find with a scan of the harness.  It is rare, I think, that using assume(a=b) in place of assignment a=b slows down the proof checking.

leegeth: Thanks Mark. I will make a follow up change to update the assignments into assumptions

cobusve (approved):

PR #12, opened 2021-03-23
Change void * in interface to Command_t **

muneebahmed10:
*Description*:
The `void *` pointers in the messaging interface will always be `Command_t **` double pointers, so this makes that change for increased readability and compile-time type checking.

### DIFF
--- a/source/include/agent_message.h
+++ b/source/include/agent_message.h
@@ -35,25 +35,29 @@
 #include <stdint.h>
 #include <stdbool.h>
 
+/* Declare here so interface functions can use. */
+struct Command;
+typedef struct Command               Command_t;
+
 /**
  * @ingroup mqtt_agent_struct_types
  * @brief Context with which tasks may deliver messages to the agent.
  */
 struct AgentMessageContext;
-typedef struct AgentMessageContext AgentMessageContext_t;
+typedef struct AgentMessageContext   AgentMessageContext_t;
 
 /**
  * @brief Send a message to the specified context.
  * Must be thread safe.
  *
  * @param[in] pMsgCtx An #AgentMessageContext_t.
- * @param[in] pData Pointer to element to send to queue.
+ * @param[in] pCommandToSend Pointer to address to send to queue.
  * @param[in] blockTimeMs Block time to wait for a send.
  *
  * @return `true` if send was successful, else `false`.
  */
 typedef bool ( * AgentMessageSend_t )( AgentMessageContext_t * pMsgCtx,
-                                       const void * pData,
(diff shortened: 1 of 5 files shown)


source/include/agent_message.h
```diff
@@ -35,39 +35,90 @@
...
  * @param[in] pMsgCtx An #AgentMessageContext_t.
- * @param[in] pData Pointer to element to send to queue.
+ * @param[in] pCommandToSend Pointer to address to send to queue.
  * @param[in] blockTimeMs Block time to wait for a send.
  *
  * @return `true` if send was successful, else `false`.
  */
 typedef bool ( * AgentMessageSend_t )( AgentMessageContext_t * pMsgCtx,
-                                       const void * pData,
+                                       Command_t * const * pCommandToSend,
```
aggarw13, line 60: This is non-readable. Could there be another typedef like `typedef Command_t * CommandAddr_t` so that this parameter type can be `const CommandAddr_t * pCommandToSend`?

test/unit-test/mqtt_agent_command_functions_utest.c
```diff
@@ -98,25 +96,17 @@ static bool stubSend( AgentMessageContext_t * pMsgCtx,
...
-    }
-    else
-    {
-        *pCommandToReceive = NULL;
-        receiveCounter++;
-        return false;
+        *pReceivedCommand = pMsgCtx->pSentCommand;
     }
+
+    return( receiveCounter++ == 0 );
```
aggarw13, line 109: Still prefer that the return value be explicitly set in the block for `receiveCounter == 0` when it already exists. If the block wouldn't have existed, I would have been fine with the expression in the return statement

source/include/agent_message.h
muneebahmed10: What's not readable about it? To me, creating another typedef just for this function that the user has to look up would be more unreadable.

test/unit-test/mqtt_agent_command_functions_utest.c
muneebahmed10: Changede

source/include/agent_message.h
aggarw13: Yes, having another typedef will add the cost of looking up another typedef but the typedef will clearly mention the meaning of the `Command_t * const *` type, which is the storage location of the Command structure.
I don't have a strong opinion against it but it definitely caused me more time to understand the type on looking at it first than I would have liked to (which could be mitigated imo with another typedef)

aggarw13 (approved):

dan4thewin (approved):

PR #19, opened 2021-03-30
[Mqtt Agent] MQTT Command Loop Complete Coverage

gshvang:
PR for 100% coverage of MQTT_Command Loop function.


test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +269,45 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
     return MQTTSuccess;
 }
 
+/**
+ * @brief A stub for MQTT_ProcessLoop function to be used to test multiple calls to MQTT_ProcessLoop.
+ */
+MQTTStatus_t MQTT_ProcessLoop_stub_multiple( MQTTContext_t * pContext,
+                                             uint32_t timeoutMs )
```
muneebahmed10, line 276: This needs an additional parameter to match CMock's stub function signature
```suggestion
                                             uint32_t timeoutMs,
                                             int NumCalls )
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +269,45 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
...
+        status = MQTTSuccess;
+    }
+    else if( processLoopCounter == 1 )
+    {
+        pMqttAgentContext->packetReceivedInLoop = true;
+        status = MQTTRecvFailed;
+    }
+    else
+    {
+        pMqttAgentContext->packetReceivedInLoop = false;
```
muneebahmed10, line 302: When will this `else` ever execute? If the loop counter is 1 then `MQTTRecvFailed` is returned and the process loop won't be called again

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1104,22 +1154,25 @@ void test_MQTTAgent_CommandLoop_with_empty_command_queue( void )
...
-    pReturnFlags.endLoop = true;
+    pReturnFlags.endLoop = false;
 
     MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
     MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &pReturnFlags );
     MQTT_ProcessLoop_IgnoreAndReturn( MQTTSuccess );
 
+    MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTBadParameter );
+    MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &pReturnFlags );
+    MQTT_ProcessLoop_IgnoreAndReturn( MQTTBadParameter );
```
muneebahmed10, line 1171: With this, lines 1165-1167 become unused. Also, we should probably be in the habit of using expect rather than ignore for failure cases or when we expect to call it exactly once/a few times, both of which apply here

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1305,6 +1354,19 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
     /* Ensure that callback is invoked. */
     TEST_ASSERT_EQUAL( 2, commandCompleteCallbackCount );
 
+    /* Invoking mqttEventCallback with MQTT_PACKET_TYPE_CONNECT packet type to cover default case. */
+    packetType = MQTT_PACKET_TYPE_CONNECT;
```
muneebahmed10, line 1358: This is literally not possible. If there's a path in the switch that can't be covered otherwise then that should be a code change.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1355,7 +1417,7 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
 
     TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
     /* Ensure that callback is invoked. */
-    TEST_ASSERT_EQUAL( 6, commandCompleteCallbackCount );
+    TEST_ASSERT_EQUAL( 7, commandCompleteCallbackCount );
```
muneebahmed10, line 1420: We should write our tests such that rearranging some or inserting one in the middle shouldn't require incrementing a variable in every assert that follows. If they're testing separate things, they should be separate tests, or this variable should be reset between API calls.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1416,5 +1478,39 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    /* Initializing command to be sent to the commandLoop. */
+    commandToSend.commandType = PUBLISH;
+    commandToSend.pCommandCompleteCallback = stubCompletionCallback;
+    commandToSend.pCmdContext = NULL;
+    commandToSend.pArgs = NULL;
+    mqttAgentContext.agentInterface.pMsgCtx->pSentCommand = &commandToSend;
+
+    mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
+
+    TEST_ASSERT_EQUAL( MQTTRecvFailed, mqttStatus );
```
muneebahmed10, line 1513: Why should this return `MQTTRecvFailed`? Everything else in this test implies this should be successful.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1416,5 +1478,39 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    setupAgentContext( &mqttAgentContext );
+    mqttAgentContext.mqttContext.connectStatus = MQTTConnected;
+    pReturnFlags.addAcknowledgment = false;
+    pReturnFlags.runProcessLoop = true;
+    pReturnFlags.endLoop = true;
+
+    MQTTAgentCommand_Publish_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTTAgentCommand_Publish_ReturnThruPtr_pReturnFlags( &pReturnFlags );
+    /* Calling MQTT_ProcessLoop multiple times. */
+    MQTT_ProcessLoop_StubWithCallback( MQTT_ProcessLoop_stub_multiple );
```
muneebahmed10, line 1502: `_StubWithCallback` is deprecated according to [CMock documentation](https://github.com/ThrowTheSwitch/CMock/blob/master/docs/CMock_Summary.md#callback). This should be `_Stub` if we haven't added any expect calls for `MQTT_ProcessLoop` and don't want CMock to keep track of the number of invocations; if we do, then it's `_AddCallback`
```suggestion
    MQTT_ProcessLoop_Stub( MQTT_ProcessLoop_stub_multiple );
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +269,45 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
...
+    pMqttAgentContext->packetReceivedInLoop = false;
+
+    if( processLoopCounter == 0 )
+    {
+        pMqttAgentContext->packetReceivedInLoop = true;
+        status = MQTTSuccess;
+    }
+    else if( processLoopCounter == 1 )
+    {
+        pMqttAgentContext->packetReceivedInLoop = true;
```
muneebahmed10, line 297: You don't need to set this flag since it's set when you call `pContext->appCallback()`. It would make more sense to just not call the callback when you want to set it to false.

muneebahmed10 (changes requested):

test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +270,37 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
...
+
+    pContext->appCallback( pContext, &packetInfo, &deserializedInfo );
+
+    if( processLoopCounter == 0 )
+    {
+        status = MQTTSuccess;
+    }
+    else
+    {
+        status = MQTTRecvFailed;
```
muneebahmed10, line 296: Comment why we're returning failure here

test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +270,37 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
...
+{
+    MQTTPacketInfo_t packetInfo;
+    MQTTDeserializedInfo_t deserializedInfo;
+    MQTTAgentContext_t * pMqttAgentContext;
+    MQTTStatus_t status;
+
+    packetInfo.type = packetType;
+    deserializedInfo.packetIdentifier = packetIdentifier;
+
+    pContext->appCallback( pContext, &packetInfo, &deserializedInfo );
```
muneebahmed10, line 288: I thought you were only going to call this based on the loop counter?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1104,22 +1147,23 @@ void test_MQTTAgent_CommandLoop_with_empty_command_queue( void )
...
+    mqttAgentContext.mqttContext.connectStatus = MQTTNotConnected;
     pReturnFlags.addAcknowledgment = false;
     pReturnFlags.runProcessLoop = true;
-    pReturnFlags.endLoop = true;
+    pReturnFlags.endLoop = false;
 
     MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
     MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &pReturnFlags );
-    MQTT_ProcessLoop_IgnoreAndReturn( MQTTSuccess );
+    MQTT_ProcessLoop_Stub( MQTT_ProcessLoop_stub );
```
muneebahmed10, line 1160: What's the point of this stub if it doesn't do anything?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1104,22 +1147,23 @@ void test_MQTTAgent_CommandLoop_with_empty_command_queue( void )
...
+
+    MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTBadParameter );
```
muneebahmed10, line 1162: It really shouldn't be returning bad parameter when all the parameters are valid. Receive failed makes more sense

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1104,22 +1147,23 @@ void test_MQTTAgent_CommandLoop_with_empty_command_queue( void )
...
 
     mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
 
-    TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
+    TEST_ASSERT_EQUAL( MQTTBadParameter, mqttStatus );
```
muneebahmed10, line 1166: What was the reason to test for failure here? A comment would be helpful

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1173,13 +1216,12 @@ void test_MQTTAgent_CommandLoop_add_acknowledgment_success( void )
...
     MQTTStatus_t mqttStatus;
     MQTTAgentContext_t mqttAgentContext;
-    MQTTAgentCommandFuncReturns_t pReturnFlags;
     Command_t commandToSend;
 
     setupAgentContext( &mqttAgentContext );
     mqttAgentContext.mqttContext.connectStatus = MQTTConnected;
     pReturnFlags.addAcknowledgment = true;
-    pReturnFlags.runProcessLoop = true;
+    pReturnFlags.runProcessLoop = false;
```
muneebahmed10, line 1224: If this is false do you need the ignore and return?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -258,6 +270,37 @@ MQTTStatus_t MQTT_ProcessLoop_stub( MQTTContext_t * pContext,
...
+    MQTTDeserializedInfo_t deserializedInfo;
+    MQTTAgentContext_t * pMqttAgentContext;
+    MQTTStatus_t status;
+
+    packetInfo.type = packetType;
+    deserializedInfo.packetIdentifier = packetIdentifier;
+
+    pContext->appCallback( pContext, &packetInfo, &deserializedInfo );
+
+    if( processLoopCounter == 0 )
```
muneebahmed10, line 290: Do we even need this when `NumCalls` exists?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1380,41 +1413,68 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
     TEST_ASSERT_EQUAL( MQTTSuccess, mqttStatus );
     /* Ensure that callback is not invoked. */
-    TEST_ASSERT_EQUAL( 9, commandCompleteCallbackCount );
+    TEST_ASSERT_EQUAL( 0, commandCompleteCallbackCount );
+}
+
+/**
+ * @brief Test MQTT_ProcessLoop failure case via calling it multiple times.
+ */
+void test_MQTT_ProcessLoop_failure( void )
```
muneebahmed10, line 1460: This name can be more descriptive as to what we're testing
```suggestion
void test_MQTTAgent_CommandLoop_ProcessLoop_failure( void )
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -93,6 +93,11 @@ static uint8_t packetType;
  */
 static uint16_t packetIdentifier;
 
+/**
+ * @brief Return flags to use for test.
+ */
+static MQTTAgentCommandFuncReturns_t pReturnFlags;
```
muneebahmed10, line 99: Should this be memset to 0 for every test?

muneebahmed10 (commented):
I believe this needs a rebase with the latest `main` branch?

test/unit-test/mqtt_agent_utest.c
gshvang: It is not a pointer. I have changed the name.

test/unit-test/mqtt_agent_utest.c
gshvang: No, its need to be called in both cases for coverage.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1102,22 +1137,21 @@ void test_MQTTAgent_CommandLoop_with_empty_command_queue( void )
...
+    mqttAgentContext.mqttContext.connectStatus = MQTTNotConnected;
+    returnFlags.addAcknowledgment = false;
+    returnFlags.runProcessLoop = true;
+    returnFlags.endLoop = false;
 
     MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
-    MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &pReturnFlags );
-    MQTT_ProcessLoop_IgnoreAndReturn( MQTTSuccess );
+    MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &returnFlags );
+    MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTRecvFailed );
```
muneebahmed10, line 1150: Looking at MQTTAgentCommand_ProcessLoop() it can't return anything except MQTTSuccess: https://github.com/FreeRTOS/coreMQTT-Agent/blob/560ce2cc9c60e6aa1be7d19a62a2d95eb2e00b54/source/mqtt_agent_command_functions.c#L55
We should probably be making sure the values we use in our unit tests are actually possible. If this is for coverage, we should be testing paths where it is actually possible for there to be an error. If we need to write another test case that's specifically for this situation then we can just do so.

test/unit-test/mqtt_agent_utest.c
muneebahmed10: It not being a pointer doesn't affect whether we should reset it or not. If we use it for multiple tests we should probably reset it.

test/unit-test/mqtt_agent_utest.c
gshvang: Added a new test case for coverage.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,194 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    setupAgentContext( &mqttAgentContext );
+
+    mqttAgentContext.mqttContext.connectStatus = MQTTNotConnected;
+    returnFlags.addAcknowledgment = false;
+    returnFlags.runProcessLoop = false;
+    returnFlags.endLoop = false;
+
+    /* Initializing command to be sent to the commandLoop. */
+    commandToSend.commandType = PUBLISH;
+    commandToSend.pCommandCompleteCallback = NULL;
```
muneebahmed10, line 1509: You should probably add the stub callback and check that it's invoked twice

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,194 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+ */
+void test_MQTTAgent_CommandLoop_with_multiple_commands( void )
+{
+    MQTTStatus_t mqttStatus;
+    MQTTAgentContext_t mqttAgentContext;
+    Command_t commandToSend = { 0 };
+
+    setupAgentContext( &mqttAgentContext );
+
+    mqttAgentContext.mqttContext.connectStatus = MQTTNotConnected;
```
muneebahmed10, line 1502: This doesn't look useful for this test
```suggestion
```

muneebahmed10 (approved):

source/mqtt_agent.c
```diff
@@ -657,7 +657,6 @@ static void mqttEventCallback( MQTTContext_t * pMqttContext,
                  * PINGRESP with the use of MQTT_ProcessLoop API function. */
                 LogWarn( ( "PINGRESP should not be handled by the application "
                            "callback when using MQTT_ProcessLoop.\n" ) );
-                break;
```
aggarg, line 660: Why is this removed?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -270,6 +278,36 @@ MQTTStatus_t MQTT_ProcessLoop_CustomStub( MQTTContext_t * pContext,
     return MQTTSuccess;
 }
 
+/**
+ * @brief A stub for MQTT_ProcessLoop function to be used to test multiple calls to MQTT_ProcessLoop.
+ */
+MQTTStatus_t MQTT_ProcessLoop_stub_multiple( MQTTContext_t * pContext,
+                                             uint32_t timeoutMs,
+                                             int NumCalls )
```
aggarg, line 286:
```suggestion
                                             int numCalls )
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,197 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    setupAgentContext( &mqttAgentContext );
+
+    mqttAgentContext.mqttContext.connectStatus = MQTTConnected;
+    returnFlags.addAcknowledgment = false;
+    returnFlags.runProcessLoop = true;
+    returnFlags.endLoop = false;
+
+    MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &returnFlags );
+    /* Calling MQTT_ProcessLoop multiple times and returning failure second time. */
```
aggarg, line 1483: I see that `MQTT_ProcessLoop_Stub` is only called once while the comment says that it is called multiple times. Does `MQTTAgent_CommandLoop` call it? If so, would you please update the comment. Else, please update the test.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,197 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+
+    mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
+
+    TEST_ASSERT_EQUAL( MQTTRecvFailed, mqttStatus );
+}
+
+/**
+ * @brief Test MQTTAgent_CommandLoop behavior processing multiple commands in command queue.
+ */
+void test_MQTTAgent_CommandLoop_with_multiple_commands( void )
```
aggarg, line 1494: What is this test checking - 1. Second command fails to get posted to the queue. OR 2. Second command gets posted to the queue and fails while processing? Would you please update the test name and comment accordingly.

source/mqtt_agent.c
gshvang: To cover the default case as it was not being covered otherwise by any other unit test case.

test/unit-test/mqtt_agent_utest.c
gshvang: Yeah MQTTAgent_CommandLoop call it. I have updated the comment.

test/unit-test/mqtt_agent_utest.c
gshvang: It is testing the second scenario. I have updated the comment.

source/mqtt_agent.c
aggarg: Will this will result in contradicting messages - One, saying that `PINGRESP` is received and should not be handled and Second, saying that unknown packet type received?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,197 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    returnFlags.addAcknowledgment = false;
+    returnFlags.runProcessLoop = true;
+    returnFlags.endLoop = false;
+
+    MQTTAgentCommand_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTTAgentCommand_ProcessLoop_ReturnThruPtr_pReturnFlags( &returnFlags );
+    /* MQTTAgent_CommandLoop calling MQTT_ProcessLoop multiple times and returning failure second time. */
+    MQTT_ProcessLoop_Stub( MQTT_ProcessLoop_stub_multiple );
+
+    mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
```
aggarg, line 1486:
```suggestion
    /* The following stub ensures that the second or later call to
     * MQTT_ProcessLoop will fail. */
    MQTT_ProcessLoop_Stub( MQTT_ProcessLoop_FailSecondAndLaterCallsStub );

    /* MQTTAgent_CommandLoop keeps calling MQTT_ProcessLoop until an error occurs.
     * In this case, the second call to MQTT_ProcessLoop will error out (as ensured
     * by the above stub) and therefore, we expect the MQTTAgent_CommandLoop
     * to fail.*/
    mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,197 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+
+    TEST_ASSERT_EQUAL( MQTTRecvFailed, mqttStatus );
+}
+
+/**
+ * @brief Test MQTTAgent_CommandLoop failure while processing multiple commands in command queue.
+ */
+void test_MQTTAgent_CommandLoop_failure_processing_multiple_commands( void )
```
aggarg, line 1494:
```suggestion
/**
 * @brief Test MQTTAgent_CommandLoop failure when executing second command in
 * the command queue fails.
 */
void test_MQTTAgent_CommandLoop_failure_executing_second_command( void )
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1299,128 +1330,197 @@ void test_MQTTAgent_CommandLoop_with_eventCallback( void )
...
+    commandToSend.pCommandCompleteCallback = stubCompletionCallback;
+    commandToSend.pCmdContext = NULL;
+    commandToSend.pArgs = NULL;
+
+    mqttAgentContext.agentInterface.pMsgCtx->pSentCommand = &commandToSend;
+
+    MQTTAgentCommand_Publish_ExpectAnyArgsAndReturn( MQTTSuccess );
+    MQTTAgentCommand_Publish_ReturnThruPtr_pReturnFlags( &returnFlags );
+    MQTT_ProcessLoop_ExpectAnyArgsAndReturn( MQTTSuccess );
+    /* PUBLISH failing while processing a second PUBLISH command. */
```
aggarg, line 1518:
```suggestion
    /* The stubReceive function ensures that the CommandLoop keeps getting the
     * same command again and again. Ensure that the second PUBLISH fails. */
```

source/mqtt_agent.c
gshvang: Yeah, will address it.

aggarg (approved):

PR #24, opened 2021-04-01
Resolve most MISRA warnings

muneebahmed10:
*Description*:
Resolves most MISRA warnings, except for:
- 8.7 (external linkage, not resolvable)
- 8.13 (adding const everywhere)
- A few 17.7 (unused return of non void).
 
There are some fixes pending for the agent that will cause some code to be rearranged and some unmodified pointers to be modified, so fixing the remaining warnings right now is premature.

Also adds a MISRA config file.

### DIFF
--- a/tools/coverity/misra.config
+++ b/tools/coverity/misra.config
@@ -1,0 +1,42 @@
+// MISRA C-2012 Rules
+
+{
+    version : "2.0",
+    standard : "c2012",
+    title: "Coverity MISRA Configuration",
+    deviations : [
+        // Disable the following rules.
+        {
+            deviation: "Directive 4.5",
+            reason: "Allow names that MISRA considers ambiguous."
+        },
+        {
+            deviation: "Directive 4.8",
+            reason: "Allow inclusion of unused types. Header files for coreMQTT, which are needed by all files, define types that are not used by the agent."
+        },
+        {
+            deviation: "Directive 4.9",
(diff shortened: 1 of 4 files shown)


tools/coverity/misra.config
```diff
@@ -0,0 +1,42 @@
...
+    title: "Coverity MISRA Configuration",
+    deviations : [
+        // Disable the following rules.
+        {
+            deviation: "Directive 4.5",
+            reason: "Allow names that MISRA considers ambiguous."
+        },
+        {
+            deviation: "Directive 4.8",
+            reason: "Allow inclusion of unused types. Header files for coreMQTT, which are needed by all files, define types that are not used by the agent."
```
leegeth, line 15:
```suggestion
            reason: "Allow inclusion of unused types. Header files for coreMQTT-Agent, which are needed by all files, define types that are not used by the agent."
```

leegeth: Please handle that for all the similar instances

muneebahmed10: No, this is correct. The MISRA warnings from removing this from the config are due to coreMQTT header files.

leegeth: Yea. Sounds good. Thanks

leegeth (approved):

n9wxu (approved):

PR #25, opened 2021-04-02
Release commands when canceled

muneebahmed10:
*Description*:

The `releaseCommand` function pointer needs to be called whenever a command has terminated, otherwise there will be a memory leak as the allocated command is never released. This adds the call to `releaseCommand` to all cases when a command is terminated, which was previously missing for error cases.

### DIFF
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -739,27 +739,23 @@ static void concludeCommand( MQTTAgentContext_t * pAgentContext,
 
     ( void ) memset( &returnInfo, 0x00, sizeof( MQTTAgentReturnInfo_t ) );
     assert( pAgentContext != NULL );
+    assert( pCommand != NULL );
 
     returnInfo.returnCode = returnCode;
     returnInfo.pSubackCodes = pSubackCodes;
 
-    if( pCommand != NULL )
+    if( pCommand->pCommandCompleteCallback != NULL )
     {
-        CommandCallback_t pCallback = pCommand->pCommandCompleteCallback;
+        pCommand->pCommandCompleteCallback( pCommand->pCmdContext, &returnInfo );
+    }
 
-        if( pCallback != NULL )
-        {
-            pCallback( pCommand->pCmdContext, &returnInfo );
-        }
+    commandReleased = pAgentContext->agentInterface.releaseCommand( pCommand );
 
-        commandReleased = pAgentContext->agentInterface.releaseCommand( pCommand );
-
-        if( !commandReleased )
-        {
-            LogError( ( "Failed to release command %p of type %d.",
-                        ( void * ) pCommand,
-                        pCommand->commandType ) );
-        }
+    if( !commandReleased )
+    {
+        LogError( ( "Failed to release command %p of type %d.",
+                    ( void * ) pCommand,
+                    pCommand->commandType ) );
     }
 }
 
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -308,6 +308,7 @@ static bool addAwaitingOperation( MQTTAgentContext_t * pAgentContext,
     AckInfo_t * pendingAcks = NULL;
 
     assert( pAgentContext != NULL );
+    assert( pCommand != NULL );
     assert( packetId != MQTT_PACKET_ID_INVALID );
     pendingAcks = pAgentContext->pPendingAcks;
 
@@ -739,6 +740,7 @@ static void concludeCommand( MQTTAgentContext_t * pAgentContext,
 
     ( void ) memset( &returnInfo, 0x00, sizeof( MQTTAgentReturnInfo_t ) );
     assert( pAgentContext != NULL );
+    assert( pAgentContext->agentInterface.releaseCommand != NULL );
     assert( pCommand != NULL );
 
     returnInfo.returnCode = returnCode;
--- a/source/mqtt_agent.c
+++ b/source/mqtt_agent.c
@@ -305,9 +305,11 @@ static bool addAwaitingOperation( MQTTAgentContext_t * pAgentContext,
 {
     size_t i = 0;
     bool ackAdded = false;
-    AckInfo_t * pendingAcks = pAgentContext->pPendingAcks;
+    AckInfo_t * pendingAcks = NULL;
 
+    assert( pAgentContext != NULL );
     assert( packetId != MQTT_PACKET_ID_INVALID );
+    pendingAcks = pAgentContext->pPendingAcks;
 
     /* Look for an unused space in the array of message IDs that are still waiting to
      * be acknowledged. */
@@ -754,7 +756,7 @@ static void concludeCommand( MQTTAgentContext_t * pAgentContext,
 
         if( !commandReleased )
         {
-            LogError( ( "Command %p of type %d not released.",
+            LogError( ( "Failed to release command %p of type %d.",
                         ( void * ) pCommand,
                         pCommand->commandType ) );
         }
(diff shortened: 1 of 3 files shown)


source/mqtt_agent.c
```diff
@@ -293,6 +307,8 @@ static bool addAwaitingOperation( MQTTAgentContext_t * pAgentContext,
     bool ackAdded = false;
     AckInfo_t * pendingAcks = pAgentContext->pPendingAcks;
 
+    assert( packetId != MQTT_PACKET_ID_INVALID );
```
aggarw13, line 310: Assert check missing for the context also?
```suggestion
    assert( pAgentContext != NULL );
    assert( packetId != MQTT_PACKET_ID_INVALID );
```

source/mqtt_agent.c
```diff
@@ -718,6 +727,42 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
...
+    bool commandReleased = false;
+    MQTTAgentReturnInfo_t returnInfo;
+
+    ( void ) memset( &returnInfo, 0x00, sizeof( MQTTAgentReturnInfo_t ) );
+    assert( pAgentContext != NULL );
+
+    returnInfo.returnCode = returnCode;
+    returnInfo.pSubackCodes = pSubackCodes;
+
+    if( pCommand != NULL )
```
aggarw13, line 744: Why is sending a NULL command acceptable to the function if the fuction's is primarily for release a command object?

source/mqtt_agent.c
```diff
@@ -718,6 +727,42 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
...
+        }
+
+        commandReleased = pAgentContext->agentInterface.releaseCommand( pCommand );
+
+        if( !commandReleased )
+        {
+            LogError( ( "Command %p of type %d not released.",
+                        ( void * ) pCommand,
+                        pCommand->commandType ) );
+        }
```
aggarw13, line 760: The `commandReleased` variable can be avoided
```suggestion
         if( pAgentContext->agentInterface.releaseCommand( pCommand ) == false )
        {
            LogError( ( "Command %p of type %d not released.",
                        ( void * ) pCommand,
                        pCommand->commandType ) );
        }
```

source/mqtt_agent.c
```diff
@@ -718,6 +727,42 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
...
+        if( pCallback != NULL )
+        {
+            pCallback( pCommand->pCmdContext, &returnInfo );
+        }
+
+        commandReleased = pAgentContext->agentInterface.releaseCommand( pCommand );
+
+        if( !commandReleased )
+        {
+            LogError( ( "Command %p of type %d not released.",
```
aggarw13, line 757: (Nit) I find it better to have the failure outcome be the subject of the statement
```suggestion
            LogError( ( "Failed to release command %p of type %d.",
```

source/mqtt_agent.c
```diff
@@ -746,6 +791,8 @@ static MQTTStatus_t resendPublishes( MQTTAgentContext_t * pMqttAgentContext )
 
             if( statusResult != MQTTSuccess )
             {
+                concludeCommand( pMqttAgentContext, pFoundAck->pOriginalCommand, statusResult, NULL );
+                ( void ) memset( pFoundAck, 0x00, sizeof( AckInfo_t ) );
                 LogError( ( "Error in resending publishes. Error code=%s\n", MQTT_Status_strerror( statusResult ) ) );
```
aggarw13, line 796: Unrelated to the PR changes but I noticed the error log and thus am suggesting an improvement. The presence of "Error" in statement is redundant/
```suggestion
                LogError( ( "Unable tp resend publishes. Error code=%s\n", MQTT_Status_strerror( statusResult ) ) );
```

source/mqtt_agent.c
muneebahmed10: This shouldn't ever be NULL. However, since the ack list is part of the agent context struct, it's visible to the user. I figured I might as well add a check here in case the user for whatever reason decided they wanted to corrupt the array.

source/mqtt_agent.c
muneebahmed10: It can, but I didn't want to because that function is both doing something and returning a status. In the past when I've combined the function call and status check in the condition, the feedback was to keep it separate as that would lead to easier debugging, as you can set a breakpoint between them. The compiler will optimize it anyway.

source/mqtt_agent.c
aggarw13: If the user happens to corrupt the array, that wouldn't be considered as the correct usage of the library, would it?  Wouldn't having an assert check somewhat deter the user from meddling with the Agent context?

aggarw13 (approved):

aggarw13 (approved):

source/mqtt_agent.c
```diff
@@ -291,7 +305,11 @@ static bool addAwaitingOperation( MQTTAgentContext_t * pAgentContext,
 {
     size_t i = 0;
     bool ackAdded = false;
-    AckInfo_t * pendingAcks = pAgentContext->pPendingAcks;
+    AckInfo_t * pendingAcks = NULL;
+
+    assert( pAgentContext != NULL );
+    assert( packetId != MQTT_PACKET_ID_INVALID );
+    pendingAcks = pAgentContext->pPendingAcks;
```
gkwicker, line 312: Should you `assert` on these conditions?

```c
assert( pendingAcks != NULL );
assert( pCommand != NULL );
```

source/mqtt_agent.c
```diff
@@ -499,7 +516,9 @@ static MQTTStatus_t processCommand( MQTTAgentContext_t * pMqttAgentContext,
 
     operationStatus = commandFunction( pMqttAgentContext, pCommandArgs, &commandOutParams );
 
-    if( ( operationStatus == MQTTSuccess ) && commandOutParams.addAcknowledgment )
+    if( ( operationStatus == MQTTSuccess ) &&
+        commandOutParams.addAcknowledgment &&
```
gkwicker, line 520: `addAcknowledgment` sounds like a `bool`, so I would make this conditional more explicit:

```c
( commandOutParams.addAcknowledgement == true )
```

source/mqtt_agent.c
```diff
@@ -699,7 +704,13 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
             {
                 /* Could not send the command to the queue so release the command
                  * structure again. */
-                pMqttAgentContext->agentInterface.releaseCommand( pCommand );
+                commandReleased = pMqttAgentContext->agentInterface.releaseCommand( pCommand );
```
gkwicker, line 707: We assert in mqtt_agent_command_functions.c:239 if `releaseCommand` isn't defined, should this function do the same? e.g.:

```c
assert( pMqttAgentContext->agentInterface.releaseCommand != NULL );
```

source/mqtt_agent.c
```diff
@@ -718,6 +729,38 @@ static MQTTStatus_t createAndAddCommand( CommandType_t commandType,
...
+
+    returnInfo.returnCode = returnCode;
+    returnInfo.pSubackCodes = pSubackCodes;
+
+    if( pCommand->pCommandCompleteCallback != NULL )
+    {
+        pCommand->pCommandCompleteCallback( pCommand->pCmdContext, &returnInfo );
+    }
+
+    commandReleased = pAgentContext->agentInterface.releaseCommand( pCommand );
```
gkwicker, line 752: We assert in mqtt_agent_command_functions.c:239 if this isn't defined, should this function do the same?

gkwicker (changes requested):
See comments

source/mqtt_agent.c
muneebahmed10: That was a deliberate choice since we haven't been using a lot of `== true` so far in this code. It's also not necessary for MISRA since coreMQTT includes stdbool.h. Does adding `== true` add any value here?

source/mqtt_agent.c
gkwicker: No need, as long as you've validated everything beforehand.

source/mqtt_agent.c
gkwicker: I'm fine with this as-is as long as it passes all static checks.  I think it's just more readable if all of the expressions in the conditional have parentheses+explicit values to compare with.

gkwicker (approved):

PR #27, opened 2021-04-06
CBMC proof for MQTTAgent_CommandLoop

leegeth:
CBMC proof for MQTTAgent_CommandLoop API.

### DIFF
--- a/test/cbmc/stubs/agent_command_functions_stub.c
+++ b/test/cbmc/stubs/agent_command_functions_stub.c
@@ -22,7 +22,7 @@
 
 /**
  * @file agent_command_functions_stub.c
- * @brief Function stub for processing an MQTT agent command.
+ * @brief Function stubs for processing an MQTT agent command.
  */
 
 /* MQTT Agent command functions include. */
@@ -34,14 +34,16 @@ static MQTTStatus_t MQTTAgentCommand_Stub( MQTTAgentContext_t * pMqttAgentContex
 {
     MQTTStatus_t returnStatus;
     uint16_t packetId;
-    MQTTAgentCommandFuncReturns_t returnFlags;
 
     __CPROVER_assert( pMqttAgentContext != NULL, "MQTT Agent context is not NULL." );
     __CPROVER_assert( pReturnFlags != NULL, "Return flags pointer is not NULL." );
 
-    *pReturnFlags = returnFlags;
+    __CPROVER_assume( packetId > 0U );
 
-    __CPROVER_assume( returnFlags.packetId > 0U );
+    pReturnFlags->packetId = packetId;
+    pReturnFlags->endLoop = nondet_bool();
+    pReturnFlags->addAcknowledgment = nondet_bool();
+    pReturnFlags->runProcessLoop = nondet_bool();
 
     return returnStatus;
 }
@@ -52,9 +54,15 @@ MQTTStatus_t MQTTAgentCommand_ProcessLoop( MQTTAgentContext_t * pMqttAgentContex
                                            void * pUnusedArg,
                                            MQTTAgentCommandFuncReturns_t * pReturnFlags )
 {
-    return MQTTAgentCommand_Stub( pMqttAgentContext,
-                                  pUnusedArg,
-                                  pReturnFlags );
+    MQTTStatus_t returnStatus;
+
+    returnStatus = MQTTAgentCommand_Stub( pMqttAgentContext,
+                                          pUnusedArg,
+                                          pReturnFlags );
+
+    pReturnFlags->runProcessLoop = false;
+
+    return returnStatus;
 }
 
 /*-----------------------------------------------------------*/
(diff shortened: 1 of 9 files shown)


test/cbmc/stubs/agent_command_functions_stub.c
```diff
@@ -0,0 +1,198 @@
...
+                                         void * pUnusedArg,
+                                         MQTTAgentCommandFuncReturns_t * pReturnFlags )
+{
+    MQTTStatus_t returnStatus;
+
+    returnStatus = MQTTAgentCommand_Stub( pMqttAgentContext,
+                                          pUnusedArg,
+                                          pReturnFlags );
+
+    pReturnFlags->endLoop = true;
```
yourslab, line 193: I noticed that the actual `MQTTAgentCommand_Terminate` actually sets all the other parameters besides `endLoop` to false as well.

test/cbmc/stubs/agent_command_functions_stub.c
```diff
@@ -0,0 +1,198 @@
...
+
+    __CPROVER_assert( pMqttAgentContext != NULL, "MQTT Agent context is not NULL." );
+    __CPROVER_assert( pReturnFlags != NULL, "Return flags pointer is not NULL." );
+
+    __CPROVER_assume( packetId > 0U );
+
+    pReturnFlags->packetId = packetId;
+    pReturnFlags->endLoop = nondet_bool();
+    pReturnFlags->addAcknowledgment = nondet_bool();
+    pReturnFlags->runProcessLoop = nondet_bool();
```
yourslab, line 46: Would it be better to just set it all to `false` by default since the actual functions would just memset the entire struct to 0?

test/cbmc/stubs/core_mqtt_stubs.c
```diff
@@ -0,0 +1,149 @@
...
+
+            __CPROVER_assume( topicNameLength < 32U );
+            pPublishInfo->pTopicName = malloc( topicNameLength );
+            __CPROVER_assume( pPublishInfo->pTopicName != NULL );
+            pPublishInfo->topicNameLength = topicNameLength;
+
+            __CPROVER_assume( payloadLength < 64U );
+            pPublishInfo->pPayload = malloc( payloadLength );
+            __CPROVER_assume( pPublishInfo->pPayload != NULL );
+            pPublishInfo->payloadLength = payloadLength;
```
yourslab, line 133: Same question from PR #28 about possibly having a `NULL` payload

test/cbmc/stubs/agent_message_stubs.c
```diff
@@ -42,7 +90,17 @@ bool AgentMessageRecvStub( AgentMessageContext_t * pMsgCtx,
...
+    bool returnStatus;
+
+    __CPROVER_assert( pMsgCtx != NULL,
+                      "MQTTAgent context is not NULL." );
+    __CPROVER_assert( pBuffer != NULL,
+                      "Command buffer is not NULL." );
+
+    command = allocateCommand();
+
+    *( ( Command_t ** ) pBuffer ) = command;
```
yourslab, line 103: Any reason for this formatting instead of using the one below?
```suggestion
    ( Command_t * ) pBuffer = command;
```

test/cbmc/stubs/core_mqtt_stubs.c
```diff
@@ -0,0 +1,149 @@
...
+        {
+            pPublishInfo = malloc( sizeof( MQTTPublishInfo_t ) );
+            __CPROVER_assume( pPublishInfo != NULL );
+
+            __CPROVER_assume( topicNameLength < 32U );
+            pPublishInfo->pTopicName = malloc( topicNameLength );
+            __CPROVER_assume( pPublishInfo->pTopicName != NULL );
+            pPublishInfo->topicNameLength = topicNameLength;
+
+            __CPROVER_assume( payloadLength < 64U );
```
yourslab, line 130: Also for the magic numbers

yourslab (approved):
Just a few minor comments again that may not need addressing. Thanks for your contributions Lee. I'll miss having you on the team.

test/cbmc/stubs/agent_command_functions_stub.c
leegeth: I think it is better to be independent to the maximum extend of how we have implemented the command functions in testing memory safety. This will give us a good confidence that any future changes in these implementations are also memory safe.

test/cbmc/stubs/agent_command_functions_stub.c
leegeth: The same response to the above question applies here as well.

test/cbmc/stubs/agent_message_stubs.c
leegeth: Yes. We will have to pass the pointer to command malloced to the pointer address contained in the pBuffer. That will not work in the latter case.

test/cbmc/stubs/core_mqtt_stubs.c
leegeth: Yes That is a good thought. I will update the PR.

test/cbmc/stubs/core_mqtt_stubs.c
leegeth: Yes. Good catch. I think an incoming publish can have NULL payload. Will address it

yourslab (approved):

aggarg (approved):

PR #28, opened 2021-04-07
CBMC proof for MQTTAgent_ResumeSession API

leegeth:
CBMC proof for MQTTAgent_ResumeSession API

PR #29, opened 2021-04-07
[MQTT Agent ]Cbmc proofs for Connect, Disconnect, Ping and ProcessLoop Command functions

gshvang:
This PR contains the cbmc proofs for the following MQTTAgentCommnad_functions:
1. MQTTAgentCommnad_Connect
2. MQTTAgentCommnad_Disconnect
3. MQTTAgentCommnad_Ping
4. MQTTAgentCommnad_ProcessLoop

### DIFF
--- a/test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
+++ b/test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
@@ -27,6 +27,7 @@
 
 /* MQTT agent include. */
 #include "mqtt_agent_command_functions.h"
+#include "mqtt_agent_cbmc_state.h"
 
 void harness()
 {
@@ -38,7 +39,7 @@ void harness()
     __CPROVER_assume( pMqttAgentContext != NULL );
     pReturnFlags = malloc( sizeof( MQTTAgentCommandFuncReturns_t ) );
     __CPROVER_assume( pReturnFlags != NULL );
-    pConnectArgs = malloc( sizeof( MQTTAgentConnectArgs_t ) );
+    pConnectArgs = allocateConnectArgs(NULL);
     __CPROVER_assume( pConnectArgs != NULL );
     MQTTAgentCommand_Connect( pMqttAgentContext, pConnectArgs, pReturnFlags );
 }
--- a/test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
+++ b/test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
@@ -1,0 +1,44 @@
+/*
+ * coreMQTT-Agent v1.0.0
+ * Copyright (C) 2020 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * @file MQTTAgentCommand_Connect_harness.c
+ * @brief Implements the proof harness for MQTTAgentCommand_Connect function.
+ */
+
+/* MQTT agent include. */
+#include "mqtt_agent_command_functions.h"
+
+void harness()
+{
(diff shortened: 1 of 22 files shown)


test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
```diff
@@ -0,0 +1,44 @@
...
+/* MQTT agent include. */
+#include "mqtt_agent_command_functions.h"
+
+void harness()
+{
+    MQTTAgentContext_t * pMqttAgentContext;
+    MQTTAgentCommandFuncReturns_t * pReturnFlags;
+    MQTTAgentConnectArgs_t * pConnectArgs;
+
+    pMqttAgentContext = malloc( sizeof( MQTTAgentContext_t ) );
```
yourslab, line 37: Is there a reason why this struct is being allocated with arbitrary memory from the heap instead of using this function to allocate it?
https://github.com/FreeRTOS/coreMQTT-Agent/blob/main/test/cbmc/sources/mqtt_agent_cbmc_state.c#L66

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
```diff
@@ -0,0 +1,44 @@
...
+    __CPROVER_assume( pMqttAgentContext != NULL );
+    pReturnFlags = malloc( sizeof( MQTTAgentCommandFuncReturns_t ) );
+    __CPROVER_assume( pReturnFlags != NULL );
+    pConnectArgs = malloc( sizeof( MQTTAgentConnectArgs_t ) );
+    __CPROVER_assume( pConnectArgs != NULL );
+    MQTTAgentCommand_Connect( pMqttAgentContext, pConnectArgs, pReturnFlags );
```
yourslab, line 43: In addition, we will also need to stub `MQTT_Connect`. That means to replace the call to the original function with another definition. The CBMC definition (or stub) doesn't need do exactly what the original function did but instead, make assertions about the parameters passed to the function or modify the parameters like how the original function did. Lee has written good examples here:
https://github.com/FreeRTOS/coreMQTT-Agent/pull/28/files#diff-6c58950d4f5c341835d2ab23348d98b95be8ab8b27f2ffa06c620a647262fd10R162

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
gshvang: I have used it because in this function we are not using the message interface and some other members of the mqttAgentContext struct here in this function, so calling that function will be an unnnecessary overhead in terms of memory allocation, so didn't used it here.

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
gshvang: I have already make an assumption that the args cannot be null. Is it still recommended to stub MQTT_connect  and verify it there?

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
```diff
@@ -0,0 +1,44 @@
...
+{
+    MQTTAgentContext_t * pMqttAgentContext;
+    MQTTAgentCommandFuncReturns_t * pReturnFlags;
+    MQTTAgentConnectArgs_t * pConnectArgs;
+
+    pMqttAgentContext = malloc( sizeof( MQTTAgentContext_t ) );
+    __CPROVER_assume( pMqttAgentContext != NULL );
+    pReturnFlags = malloc( sizeof( MQTTAgentCommandFuncReturns_t ) );
+    __CPROVER_assume( pReturnFlags != NULL );
+    pConnectArgs = malloc( sizeof( MQTTAgentConnectArgs_t ) );
```
yourslab, line 41: Same here. The purpose of CBMC is to define properties with `__CPROVER_assume` about input parameters given to the function under test. By simply calling `malloc`, the only property we are stating here is that `pConnectArgs` contains memory on the heap with the same size as `MQTTAgentConnectArgs_t`. That means that each member of the struct can take on any value, but do we really want that to be the case? Sure, `sessionPresent` can take on 'true' or 'false'. The timeout can also take on any value. However, do we expect `MQTTAgentCommand_Connect` to be called with `pConnectInfo` pointing to any random address? No, we only expect  `MQTTAgentCommand_Connect` to be called when `pConnectInfo` points to a `MQTTConnectInfo_t` structure.

In summary, you need to create an `allocateConnectArgs( MQTTAgentConnectArgs_t * pConnectArgs )` function that will set `MQTTAgentConnectArgs_t` to values that we expect its caller (`MQTTAgentCommand_Connect`) to see. This applies to everything we do in CBMC.

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
```diff
@@ -0,0 +1,44 @@
...
+    __CPROVER_assume( pConnectArgs != NULL );
+    MQTTAgentCommand_Connect( pMqttAgentContext, pConnectArgs, pReturnFlags );
+}
```
yourslab, line 44: After `MQTTAgentCommand_Connect` returns, we can also `__CPROVER_assert` that `pReturnFlags` has set everything to false.

yourslab: Although I suppose the unit tests already do that for us

yourslab (changes requested):
Same comments apply to all the proofs. You'll need stubs for `MQTT_Disconnect`, `MQTT_Ping`, etc. Some have already been written by Lee. Don't forget to write allocators and validators in the `sources/mqtt_agent_cbmc_state.c` file.

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
gshvang: Yeah, unit test are already checking these

test/cbmc/proofs/MQTTAgentCommand_Connect/MQTTAgentCommand_Connect_harness.c
yourslab: Yes, you will need to stub it. Works similarly for the unit tests. We are mocking functions called by the function under test to avoid coupling.

yourslab (approved):

aggarg (approved):

PR #34, opened 2021-04-13
Add GitHub Actions workflow to automate release

aggarw13:
**Add a GitHub Action job to trigger a release process** which includes operations of: 
1. Deleting an existing release and tag, if the same tag is being re-released.
1. Creating and pushing the release tag to the repository 
1. Verifying the pushed tag (performing a diff with the commit ID which is tagged) 
1. Creating a ZIP for the release asset. 
1. Verifying the ZIP by performing a diff check and running unit tests on the unzipped files. 5. Creating a release on the repository for the tag, and uploading the ZIP as the release asset

The workflow can be manually triggered and takes the input values of **Commit ID** (to create a release for) and the **Version string** for tagging the release with

**Testing**
Here are example runs of the release job on my fork repository that created the [tag](https://github.com/aggarw13/coreMQTT-Agent/tree/v1.0.0) and [release](https://github.com/aggarw13/coreMQTT-Agent/releases/tag/v1.0.0): 
* Releasing the `v1.0.0` tag - https://github.com/aggarw13/coreMQTT-Agent/actions/runs/745933055
* Re-releasing the same tag - https://github.com/aggarw13/coreMQTT-Agent/actions/runs/746241995

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,149 @@
+name: Release automation
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_id:
+        description: 'Commit ID to tag and create a release for'
+        required: true
+      version_number:
+        description: 'Release Version Number (Eg, v1.0.0)'
+        required: true
+      delete_existing_tag_release:
+        description: 'Is this a re-release of existing tag/release? (Default: false)'
+        default: 'false'
+        required: false
+jobs:
+  clean-existing-tag-and-release:
+    if: ${{ github.event.inputs.delete_existing_tag_release == 'true' }}
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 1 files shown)


yourslab (approved):
Is there a way to make it so that this workflow becomes a part of the shared actions repo? It looks exactly the same as what exists in coreMQTT and other spoke repos.

aggarw13:
>Is there a way to make it so that this workflow becomes a part of the shared actions repo? It looks exactly the same as what exists in coreMQTT and other spoke repos.

That's a valid question. The issue is that GitHub Actions cannot contain logic to invoke another action, so the functionality of uploading artifacts cannot be commonized; though the functionality of Zipping and uploading tag can be.

aggarg (approved):

PR #37, opened 2021-04-14
Add doxygen files

muneebahmed10:
*Description*:
Adds files required to generate doxygen documentation. Should be merged after #12 as that will cause some rearrangements where types are defined.


docs/doxygen/pages.dox
```diff
@@ -0,0 +1,98 @@
+/**
+@mainpage Overview
+@anchor mqtt
+@brief Thread safe MQTT 3.1.1 client
+
+coreMQTT is an MIT licensed open source C MQTT client library for microcontroller and small microprocessor based IoT devices. It’s design is intentionally simple to ensure it has no dependency on any other library or operating system, and to better enable static analysis including memory safety proofs. That simplicity and lack of operating system dependency (coreMQTT does not require multithreading at all) means coreMQTT does not build thread safety directly into its implementation. Instead thread safety must be provided by higher level software. This web page demonstrates a coreMQTT extension that provides that higher level functionality in the form of an MQTT agent (or MQTT daemon). While the implementation demonstrated here is currently specific to FreeRTOS, there are not many dependencies on FreeRTOS, meaning the implementation can easily be adapted for use with other operating systems.
```
aggarw13, line 6: With initial description about just the coreMQTT library, the emphasis on coreMQTT Agent could be missing for a new reader. Would it be possible to highlight about MQTT Agent and separately have a section about why a newer library on top of coreMQTT library has been created? Basically, something of this structure?

**Overview**

_Description about MQTT Agent which says that it is based on the coreMQTT library and provides thread-safe implementation of an MQTT client on the FreeRTOS operating system_

**Why is there a higher level library based besides coreMQTT?**

_Brief description about coreMQTT, its benefits and its limitations_

docs/doxygen/config.doxyfile
```diff
@@ -0,0 +1,2567 @@
...
+
+DOXYFILE_ENCODING      = UTF-8
+
+# The PROJECT_NAME tag is a single word (or a sequence of words surrounded by
+# double-quotes, unless you are using Doxywizard) that should identify the
+# project for which the documentation is generated. This name is used in the
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "MQTT Agent"
```
aggarw13, line 35: If the name of the library/product is coreMQTT-Agent, should the `core` prefix be added?
```suggestion
PROJECT_NAME           = "coreMQTT Agent"
```

docs/doxygen/pages.dox
```diff
@@ -0,0 +1,98 @@
+/**
+@mainpage Overview
+@anchor mqtt
+@brief Thread safe MQTT 3.1.1 client
```
aggarw13, line 4: As this library is specific to FreeRTOS, would it be better to mention it clearly in the brief, so that customers can notice that coreMQTT is platform agnostic whereas the MQTT Agent is a thread-safe client for FreeRTOS
```suggestion
@brief Thread safe MQTT 3.1.1 client for FreeRTOS
```

source/include/agent_message.h
```diff
@@ -33,13 +33,13 @@
 
 /* Declare here so interface functions can use. */
 struct Command;
+struct AgentMessageContext;
```
aggarw13, line 36: Why was this forward declaration moved above?

aggarw13: If the goal is the group the forward declarations, should there be a `@brief` for the `Command_t` type?

source/include/mqtt_agent.h
```diff
@@ -156,15 +169,15 @@ typedef void (* IncomingPublishCallback_t )( MQTTAgentContext_t * pMqttAgentCont
  * MQTTAgent_Init(), and every API function will accept a pointer to the
  * initalized struct.
  */
-struct MQTTAgentContext
+typedef struct MQTTAgentContext
```
aggarw13, line 172: Can this definition be moved before the definition of the `IncomingPublishCallback_t` type so that the parameter can take the `MqttAgentContext_t` type?

source/include/mqtt_agent_command_functions.h
```diff
@@ -127,7 +127,7 @@ MQTTStatus_t MQTTAgentCommand_Publish( MQTTAgentContext_t * pMqttAgentContext,
  * @brief Function to execute for a SUBSCRIBE command.
  *
  * @param[in] pMqttAgentContext MQTT Agent context information.
- * @param[in] pSubscribeArgs Arguments for MQTT_Subscribe().
+ * @param[in] pVoidSubscribeArgs Arguments for MQTT_Subscribe().
```
aggarw13, line 130: Should these (throughout the file) reference the doxygen tag for the coreMQTT API function?
```suggestion
 * @param[in] pVoidSubscribeArgs Arguments for @ref MQTT_Subscribe().
```

aggarw13: Also, why the presence of `Void` in the parameter name?

aggarw13 (commented):
Can you also add the doxygen check in the CI workflow as part of this PR?

docs/doxygen/pages.dox
muneebahmed10: This library doesn't contain any FreeRTOS specific code any more, so I removed that from the description

source/include/agent_message.h
muneebahmed10: Yes, it was to group them. Command_t doesn't have a `@brief` here since it has one where the struct is defined in `mqtt_agent.h`. However, it causes warnings in doxygen to define the type and the struct in different files that I haven't figured out yet. One solution could be to move the typedef to the other file, and change everything here to `struct Command`. I left that for the next PR

source/include/mqtt_agent.h
muneebahmed10: No, since `MQTTAgentContext_t` includes `IncomingPublishCallback_t`

source/include/mqtt_agent_command_functions.h
muneebahmed10: No, since we haven't figured out yet if we want to keep linking to coreMQTT or remove those references.

The reason for void is to easily distinguish between variables when we cast them to the correct pointer type, since these parameters are passed as `void *`. I updated `MQTTAgentCommand_Connect` to also use this naming, but not `MQTTAgentCommand_Publish` since that one doesn't use a custom struct that needs to be dereferenced here, it can just be cast when passing to MQTT_Publish()

muneebahmed10:
> Can you also add the doxygen check in the CI workflow as part of this PR?

If I did, the check would not pass due to the references to coreMQTT. I left that for another PR

docs/doxygen/config.doxyfile
muneebahmed10: Thanks, updated

docs/doxygen/pages.dox
muneebahmed10: I put them in separate sections

source/include/agent_message.h
aggarw13: I think that from a reader's perpsective, it is better to be consistent about use of `Command_t` everywhere instead of using `Command_t` in one header file and `struct Command` in another API file. 
Currently, I see that the `mqtt_agent.h` file depends on `agent_message.h` to receive the `Command_t` typedef which can also be confusing to the reader because `struct Command` is defined in `mqtt_agent.h` but the file uses its typedef in the API that it sources from `agent_message.h` file, which may not be obvious.

One solution can be to define a `mqtt_agent_types.h` file where all the types are housed instead of having the structs and their typedefs spread in different files. It can be considered for a different PR though

docs/doxygen/pages.dox
```diff
@@ -0,0 +1,101 @@
...
+
+The coreMQTT Agent is a thread-safe library to serialize calls to coreMQTT, to be executed by a single thread. It provides APIs for a single, dedicated agent task to process coreMQTT related commands, and the APIs for other tasks to enqueue these commands for processing.
+
+@section coreMQTT_brief Why is there a higher level library based on coreMQTT?
+coreMQTT is an MIT licensed open source C MQTT client library for microcontroller and small microprocessor based IoT devices. Its design is intentionally simple to ensure it has no dependency on any other library or operating system, and to better enable static analysis including memory safety proofs. That simplicity and lack of operating system dependency (coreMQTT does not require multithreading at all) means coreMQTT does not build thread safety directly into its implementation. Instead thread safety must be provided by higher level software. This library is a coreMQTT extension that provides that higher level functionality in the form of an MQTT agent (or MQTT daemon).
```
aggarw13, line 9:
```suggestion
coreMQTT is an MIT licensed open source C MQTT client library for microcontroller and small microprocessor based IoT devices. Its design is intentionally simple to ensure it has no dependency on any other library or operating system, and to better enable static analysis including memory safety proofs. That simplicity and lack of operating system dependency (coreMQTT does not require multithreading at all) means coreMQTT does not build thread safety directly into its implementation. Instead, thread safety must be provided by higher level software. The coreMQTT Agent library is a coreMQTT extension that provides that higher level functionality in the form of an MQTT agent (or MQTT daemon).
```

source/include/mqtt_agent_command_functions.h
aggarw13: Why does that distinguishing matter from an API perspective? Isn't that an implementation detail?

muneebahmed10: (This function isn't even part of the API, I don't think variable naming matters too much)

aggarw13: I don't see how much value is `Void` naming pattern is helping for the implementation to be conveyed about the parameter handling. If the information has to be conveyed to the command function implementation, I think it should be conveyed in a stronger way (either at least through doxygen documentation Or through a separate parameter mentioning the type). 
However, I prefer that this information be encapsulated within the implementation than be shown in the header file.

aggarw13 (approved):

n9wxu (approved):

PR #39, opened 2021-04-16
Handle edge cases for list of pending acknowledgements

aggarw13:
Handle the following 2 edge cases for entries in the list of pending acknowledgments (or `MQTTAgentContext_t.pPendingAcks`): 
1. When an MQTT session is resumed, but the previous connection of the session had un-acknowledged SUBSCRIBE and UNSUBSCIBE requests, the list needs to be cleared of entries for those operations. This PR updates the `MQTT_ResumeSession` API for the clean-up operation.
2. In the unlikely case when an operation is unkacknowledged, and there occur 65535 operations after it that cause the packet ID value to wrap around, then the 65536th operation, if needing acknowledgement, will have the same packet ID as the still packet ID still pending acknowledgment in the list (packet ID is represented as a 16 bit integer). This PR handles this case by returning an `MQTTStateCollision` error in that case.


source/mqtt_agent.c
```diff
@@ -295,35 +305,65 @@ static bool isSpaceInPendingAckList( const MQTTAgentContext_t * pAgentContext )
...
+             * the list.
+             * Note: This is an unlikely edge case which represents that a packet ID
+             * didn't receive acknowledgement, but subsequent SUBSCRIBE/PUBLISH operations
+             * representing 65535 packet IDs were successful that caused the bit packet
+             * ID value to wrap around and reached the same packet ID as that was still
+             * pending acknowledgement.
+             */
+            status = MQTTStateCollision;
+            LogError( ( "Failed to add operation to list of pending acknowledgements: Existing entry found :"
+                        "for same packet: PacketId=%u\n", packetId ) );
```
muneebahmed10, line 348: We've been dropping the 'e' everywhere else in this repo. to match the most common American spelling.
```suggestion
            /* Check whether there exists a duplicate entry for pending
             * acknowledgment for the same packet ID that we want to add to
             * the list.
             * Note: This is an unlikely edge case which represents that a packet ID
             * didn't receive acknowledgment, but subsequent SUBSCRIBE/PUBLISH operations
             * representing 65535 packet IDs were successful that caused the bit packet
             * ID value to wrap around and reached the same packet ID as that was still
             * pending acknowledgment.
             */
            status = MQTTStateCollision;
            LogError( ( "Failed to add operation to list of pending acknowledgments: Existing entry found :"
                        "for same packet: PacketId=%u\n", packetId ) );
```

source/mqtt_agent.c
```diff
@@ -295,35 +305,65 @@ static bool isSpaceInPendingAckList( const MQTTAgentContext_t * pAgentContext )
...
+        if( ( unusedPos == MQTT_AGENT_MAX_OUTSTANDING_ACKS ) &&
+            ( pendingAcks[ i ].packetId == MQTT_PACKET_ID_INVALID ) )
         {
-            pendingAcks[ i ].packetId = packetId;
-            pendingAcks[ i ].pOriginalCommand = pCommand;
-            ackAdded = true;
+            unusedPos = i;
+            status = MQTTSuccess;
+        }
+        else if( pendingAcks[ i ].packetId == packetId )
```
muneebahmed10, line 335: Please either add an empty `else` block for MISRA, or change this `else if` to an `if`

source/mqtt_agent.c
```diff
@@ -295,35 +305,65 @@ static bool isSpaceInPendingAckList( const MQTTAgentContext_t * pAgentContext )
...
+    if( status == MQTTSuccess )
+    {
+        pendingAcks[ unusedPos ].packetId = packetId;
+        pendingAcks[ unusedPos ].pOriginalCommand = pCommand;
+    }
+    else if( status == MQTTNoMemory )
+    {
+        LogError( ( "Failed to add operation to list of pending acknowledgements: No memory available:"
+                    "PacketId=%u\n", packetId ) );
+    }
```
muneebahmed10, line 364:
```suggestion
        LogError( ( "Failed to add operation to list of pending acknowledgments: No memory available:"
                    "PacketId=%u\n", packetId ) );
    }
    else
    {
        /* Empty else MISRA 15.7 */
    }
```

source/mqtt_agent.c
```diff
@@ -515,18 +555,8 @@ static MQTTStatus_t processCommand( MQTTAgentContext_t * pMqttAgentContext,
...
-        if( !ackAdded )
-        {
-            LogError( ( "No memory to wait for acknowledgment for packet %u\n", commandOutParams.packetId ) );
-
-            /* All operations that can wait for acks (publish, subscribe,
-             * unsubscribe) require a context. */
-            operationStatus = MQTTNoMemory;
-        }
+        operationStatus = addAwaitingOperation( pMqttAgentContext, commandOutParams.packetId, pCommand );
+        ackAdded = ( operationStatus == MQTTSuccess ) ? true : false;
```
muneebahmed10, line 559: The last part isn't really necessary
```suggestion
        ackAdded = ( operationStatus == MQTTSuccess );
```

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1267,20 +1308,40 @@ void test_MQTTAgent_CommandLoop_add_acknowledgment_failure( void )
...
-
+    /* Call API under test. */
     mqttStatus = MQTTAgent_CommandLoop( &mqttAgentContext );
 
     TEST_ASSERT_EQUAL( MQTTNoMemory, mqttStatus );
 
     /* Ensure that callback is invoked. */
     TEST_ASSERT_EQUAL( 1, commandCompleteCallbackCount );
+
+    MQTTAgentCommand_Publish_ExpectAnyArgsAndReturn( MQTTSuccess );
```
muneebahmed10, line 1328: Can you reset the value of `commandCompleteCallbackCount` right here? It looks like the only reason it should be 2 later is that a test was run once before.

source/mqtt_agent.c
```diff
@@ -814,11 +845,23 @@ static void clearPendingAcknowledgments( MQTTAgentContext_t * pMqttAgentContext
     {
         if( pendingAcks[ i ].packetId != MQTT_PACKET_ID_INVALID )
         {
-            /* Receive failed to indicate network error. */
-            concludeCommand( pMqttAgentContext, pendingAcks[ i ].pOriginalCommand, MQTTRecvFailed, NULL );
+            bool clearEntry = true;
+
+            if( clearOnlySubUnsubEntries &&
```
muneebahmed10, line 850: Nit: Perhaps we should `assert( pendingAcks[ i ].pOriginalCommand != NULL );` before dereferencing it.

source/mqtt_agent.c
```diff
@@ -814,11 +845,23 @@ static void clearPendingAcknowledgments( MQTTAgentContext_t * pMqttAgentContext
     {
         if( pendingAcks[ i ].packetId != MQTT_PACKET_ID_INVALID )
         {
-            /* Receive failed to indicate network error. */
-            concludeCommand( pMqttAgentContext, pendingAcks[ i ].pOriginalCommand, MQTTRecvFailed, NULL );
+            bool clearEntry = true;
+
```
aggarw13, line 849:
```suggestion
            bool clearEntry = true;
            
            assert( pendingAcks[ i ].pOriginalCommand != NULL );
                        
```

test/unit-test/mqtt_agent_utest.c
aggarw13: Sure, done.

test/unit-test/mqtt_agent_utest.c
```diff
@@ -529,16 +529,57 @@ void test_MQTTAgent_ResumeSession_session_present_no_publish_found( void )
...
+
+    MQTT_PublishToResend_IgnoreAndReturn( MQTT_PACKET_ID_INVALID );
+
+    setupAgentContext( &mqttAgentContext );
+
+    /* Setup the pending ack list to contain operations for both SUBSCRIBE
+     * and UNSUBSCRIBE operations. */
+    mqttAgentContext.pPendingAcks[ 0 ].packetId = 1U;
+    mqttAgentContext.pPendingAcks[ 0 ].pOriginalCommand = &subscribeCommand;
+
```
muneebahmed10, line 564: Should we also add a PUBLISH command and test if that does *not* get cleared?

test/unit-test/mqtt_agent_utest.c
```diff
@@ -1239,23 +1280,24 @@ void test_MQTTAgent_CommandLoop_add_acknowledgment_success( void )
...
+ * cannot be added in the list of pending acknowledgments, either due to the array
+ * being full OR there being an existing entry for the same packet ID.
  */
 void test_MQTTAgent_CommandLoop_add_acknowledgment_failure( void )
 {
     MQTTStatus_t mqttStatus;
     MQTTAgentContext_t mqttAgentContext;
     size_t i = 0;
     Command_t commandToSend = { 0 };
+    uint16_t testPacketId = 1U;
```
muneebahmed10, line 1293: Could make this `const` if its value doesn't change
```suggestion
    const uint16_t testPacketId = 1U;
```

muneebahmed10 (approved):

test/unit-test/mqtt_agent_utest.c
aggarw13: Added

n9wxu (approved):

PR #40, opened 2021-04-19
Add doxygen GA and link to MQTT documentation

muneebahmed10:
*Description*:
Adds a GitHub Actions check for doxygen documentation, and links to external MQTT documentation. Since the `Command_t` typedef and struct definition are in separate files, doxygen has issues resolving references to it. This works around it by referring to the struct itself in our docs instead of the typedef.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,14 @@ jobs:
         uses: FreeRTOS/CI-CD-Github-Actions/complexity@main
         with:
           path: ./
+  doxygen:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run doxygen build
+        uses: FreeRTOS/CI-CD-Github-Actions/doxygen@main
+        with:
+          path: ./
   spell-check:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
           # We don't need to generate the coreMQTT docs, we only need the tag file. We can just download it.
           mkdir -p source/dependency/coreMQTT/docs/doxygen/output
           wget -O source/dependency/coreMQTT/docs/doxygen/output/mqtt.tag \
-          "https://docs.aws.amazon.com/embedded-csdk/202103.00/lib-ref/libraries/standard/coreMQTT/docs/doxygen/output/mqtt.tag"
+          "https://freertos.org/Documentation/api-ref/coreMQTT/docs/doxygen/output/mqtt.tag"
       - name: Run doxygen build
         uses: FreeRTOS/CI-CD-Github-Actions/doxygen@main
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Download MQTT tag
+        run: |
+          # We don't need to generate the coreMQTT docs, we only need the tag file. We can just download it.
+          mkdir -p source/dependency/coreMQTT/docs/doxygen/output
+          wget -O source/dependency/coreMQTT/docs/doxygen/output/mqtt.tag \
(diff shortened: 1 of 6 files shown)


docs/doxygen/config.doxyfile
```diff
@@ -2226,7 +2226,7 @@ SKIP_FUNCTION_MACROS   = YES
 # the path). If a tag file is not located in the directory in which doxygen is
 # run, you must also specify the path to the tagfile here.
 
-TAGFILES               =
+TAGFILES               = source/dependency/coreMQTT/docs/doxygen/output/mqtt.tag=https://docs.aws.amazon.com/embedded-csdk/202103.00/lib-ref/libraries/standard/coreMQTT/docs/doxygen/output/html
```
aggarw13, line 2229: Would be better to link to the doxygen hosted on freertos.org
```suggestion
TAGFILES               = source/dependency/coreMQTT/docs/doxygen/output/mqtt.tag=https://freertos.org/Documentation/api-ref/coreMQTT/docs/doxygen/output/html`

File: source/include/mqtt_agent_command_functions.h [hunk 2]
```diff
@@ -195,7 +195,7 @@ MQTTStatus_t MQTTAgentCommand_Ping( MQTTAgentContext_t * pMqttAgentContext,
  * @param[in] pUnusedArg Unused NULL argument.
  * @param[out] pReturnFlags Flags set to indicate actions the MQTT agent should take.
  *
- * @return `MQTTSuccess`.
+ * @return #MQTTSuccess.
```
aggarw13, line 198: How about references to the coreMQTT API functions (like `MQTT_Ping` in `MQTTAgentCommand_Ping` function documentation)?

muneebahmed10: Doxygen's smart enough to automatically link functions and data types without needing the `#`. It's just for specific enum values like this we need it

.github/workflows/ci.yml
```diff
@@ -47,6 +47,20 @@ jobs:
         uses: FreeRTOS/CI-CD-Github-Actions/complexity@main
         with:
           path: ./
+  doxygen:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Download MQTT tag
+        run: |
+          # We don't need to generate the coreMQTT docs, we only need the tag file. We can just download it.
```
aggarw13, line 56: There can be issue in the future when changes to coreMQTT do not match the documentation present in a previous release. Would it be better to just build the tag off of the coreMQTT commit submoduled in the repository instead?

.github/workflows/ci.yml
```diff
@@ -47,6 +47,20 @@ jobs:
...
+          mkdir -p source/dependency/coreMQTT/docs/doxygen/output
+          wget -O source/dependency/coreMQTT/docs/doxygen/output/mqtt.tag \
+          "https://docs.aws.amazon.com/embedded-csdk/202103.00/lib-ref/libraries/standard/coreMQTT/docs/doxygen/output/mqtt.tag"
```
aggarw13, line 59: Why is it required to download the mqtt.tag when `config.doxyfile` is configured to use the tag hosted on AWS?

muneebahmed10: The tag file can't be a URL, it has to be a file

.github/workflows/ci.yml
muneebahmed10: I disagree. It would make sense for the tag file to match the documentation it was generated with; if there's a coreMQTT update, I don't think building the documentation from the coreMQTT commit would help since it's still linking to the external docs.

docs/doxygen/config.doxyfile
muneebahmed10: Sure

.github/workflows/ci.yml
aggarw13: Wouldn't changes to coreMQTT still work as long as they are re-hosted in the subsequent release?

aggarw13 (approved):

.github/workflows/ci.yml
muneebahmed10: Yes, but if the docs are re-hosted, then the tag file from the URL will also be up to date since the tag file is part of the generated docs.

n9wxu (approved):

PR #50, opened 2021-04-28
Change void to MQTTAgentCommandContext_t

muneebahmed10:
*Description*:
The first parameter of the command completion callback is the command context pointer, which does not need to be `void *` and can be `MQTTAgentCommandContext_t *`.

### DIFF
--- a/source/include/core_mqtt_agent.h
+++ b/source/include/core_mqtt_agent.h
@@ -118,7 +118,7 @@ typedef struct MQTTAgentCommandContext MQTTAgentCommandContext_t;
  * application wants to enqueue command(s) with non-zero blocking time, the
  * callback can notify a different task to enqueue command(s) to the MQTT agent.
  */
-typedef void (* MQTTAgentCommandCallback_t )( void * pCmdCallbackContext,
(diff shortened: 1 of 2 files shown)


source/include/core_mqtt_agent.h
```diff
@@ -118,7 +118,7 @@ typedef struct MQTTAgentCommandContext MQTTAgentCommandContext_t;
  * application wants to enqueue command(s) with non-zero blocking time, the
  * callback can notify a different task to enqueue command(s) to the MQTT agent.
  */
-typedef void (* MQTTAgentCommandCallback_t )( void * pCmdCallbackContext,
+typedef void (* MQTTAgentCommandCallback_t )( MQTTAgentCommandContext_t * pCmdCallbackContext,
```
aggarw13, line 121: Can you also update the doxygen examples for the type?

muneebahmed10: Are they not updated here?

aggarw13: Oops, I didn't notice that most of the changes were related to doxygen updates. 😅

aggarw13 (approved):

dan4thewin (approved):

PR #53, opened 2021-06-10
Add extern "C" for C linkage in C++

MustafaYassin:
Added extern "C" to be linkable in C++ compilation.

### DIFF
--- a/source/include/core_mqtt_agent.h
+++ b/source/include/core_mqtt_agent.h
@@ -27,6 +27,10 @@
 #ifndef CORE_MQTT_AGENT_H
 #define CORE_MQTT_AGENT_H
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 /* MQTT library includes. */
 #include "core_mqtt.h"
 #include "core_mqtt_state.h"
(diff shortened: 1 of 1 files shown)


boraozgen:
Why not use extern "C" when including?

armandas:
> Why not use extern "C" when including?

I believe it's a common sense for C libraries to add this to their headers. FreeRTOS kernel does it: https://github.com/FreeRTOS/FreeRTOS-Kernel/blob/6a84f2c1da0f8d9fd1946caa55a336ebacfeb392/include/FreeRTOS.h#L54

The benefits are:

- Library works seamlessly with C and C++.
- User don't have to waste time debugging cryptic linker errors.
- Maintainers don't have to waste time telling the users to "wrap the include in `extern "C"`.

MustafaYassin:
The benefits include: 
- cleaner user code.
- alignment with all other FreeRTOS headers.

muneebahmed10:
Hi @MustafaYassin,

Thank you for bringing this to our attention. We have decided to include this change. However, the preprocessor guards need to be surrounded by `/* *INDENT-OFF* */` / `/* *INDENT-ON* */` in order to be accepted by the uncrustify tool we use PR code checks, and there are other header files in this repository that should be updated with this as well. To save you the trouble of updating this pull request, we have already included this in #56.

Thanks again!

PR #68, opened 2021-12-01
Loop invariant update

angelonakos:
Identical to #67


angelonakos:
CI succeeded for both #67 and #68. Closing in favor of that PR.

PR #75, opened 2022-09-28
Update MQTT Agent to work with latest version of coreMQTT

AniruddhaKanhere:


source/core_mqtt_agent.c
```diff
@@ -546,8 +560,16 @@ static MQTTStatus_t processCommand( MQTTAgentContext_t * pMqttAgentContext,
...
         assert( pCommand->commandType < NUM_COMMANDS );
-        commandFunction = pCommandFunctionTable[ pCommand->commandType ];
-        pCommandArgs = pCommand->pArgs;
+
+        if( ( pCommand->commandType >= NONE ) && ( pCommand->commandType < NUM_COMMANDS ) )
+        {
+            commandFunction = pCommandFunctionTable[ pCommand->commandType ];
+            pCommandArgs = pCommand->pArgs;
+        }
+        else
```
jasonpcarroll, line 569: should an error be reported here? wouldn't the command being out of bounds of the command types indicate something faulty happened?

source/core_mqtt_agent.c
```diff
@@ -621,9 +643,9 @@ static void handleAcks( const MQTTAgentContext_t * pAgentContext,
 
 static MQTTAgentContext_t * getAgentFromMQTTContext( MQTTContext_t * pMQTTContext )
 {
-    void * ret = pMQTTContext;
+    size_t offset = ( ( size_t ) &( ( ( MQTTAgentContext_t * ) 0 )->mqttContext ) );
```
jasonpcarroll, line 646: I see what you are doing here - cool way to portably determine the correct struct offset. im just worried that this will not be misra compliant - but could probably be fixed when we make this lib misra compliant at some point.

source/core_mqtt_agent.c
```diff
@@ -980,6 +1002,15 @@ MQTTStatus_t MQTTAgent_Init( MQTTAgentContext_t * pMqttAgentContext,
                                   mqttEventCallback,
                                   pNetworkBuffer );
 
+        if( returnStatus == MQTTSuccess )
+        {
+            returnStatus = MQTT_InitStatefulQoS( &( pMqttAgentContext->mqttContext ),
```
jasonpcarroll, line 1007: would if users didnt need QoS > 0? i suppose giving them that option may mean an API change... should be able to add this to the config tho, no?

source/core_mqtt_agent.c
AniruddhaKanhere: I added a warning message.

source/core_mqtt_agent.c
AniruddhaKanhere: The earlier was not only a violation to MISRA rules, it was also a possible memory corruption issue in case anyone decided to modify the MQTT-Agent structure.

source/include/core_mqtt_agent_default_logging.h
```diff
@@ -0,0 +1,129 @@
+/*
```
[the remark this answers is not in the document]
AniruddhaKanhere: I added a config file too! :)

source/core_mqtt_agent.c
AniruddhaKanhere: Yes, good point. Since we now have a config file, I added a config macro for this too.

source/include/core_mqtt_agent_config_defaults.h
```diff
@@ -0,0 +1,119 @@
...
+/* *INDENT-OFF* */
+#ifdef __cplusplus
+    extern "C" {
+#endif
+/* *INDENT-ON* */
+
+/* MQTT_AGENT_USE_CUSTOM_CONFIG allows building the MQTT library
+ * without a custom config. If a custom config is provided, the
+ * MQTT_AGENT_USE_CUSTOM_CONFIG macro should be defined. */
+#ifdef MQTT_AGENT_USE_CUSTOM_CONFIG
```
jasonpcarroll, line 47: For consistency sake - should this not be MQTT_AGENT_DO_NOT_USE_CUSTOM_CONFIG https://github.com/FreeRTOS/coreMQTT/blob/10d85cb9f96f680072c474bef415fc76bf673642/source/include/core_mqtt_config_defaults.h#L47

source/include/core_mqtt_agent_config_defaults.h
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * @brief Whether the agent should configure the coreMQTT library to be used with publishes
+ * greater than QoS0. Setting this to 0 will disallow the coreMQTT library to send publishes
+ * with QoS > 0.
+ *
+ * <b>Possible values:</b> 0 or 1 <br>
+ * <b>Default value:</b> `1`
+ */
+#ifndef MQTT_AGENT_USE_QOS_1_2_PUBLISH
+    #define MQTT_AGENT_USE_QOS_1_2_PUBLISH    ( 1 )
```
jasonpcarroll, line 110: bit of a nit pick but should this be 1U?
```suggestion
    #define MQTT_AGENT_USE_QOS_1_2_PUBLISH    ( 1U )
```

source/include/core_mqtt_agent_default_logging.h
```diff
@@ -0,0 +1,129 @@
...
+ * coreMQTT-Agent v1.1.0
+ * Copyright (C) 2022 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ *
```
jasonpcarroll, line 4: Should this include core_mqtt_agent_config_default if a user chooses to implement their own logging? i can foresee a file that includes this first then core_mqtt_agent_config_default then core_mqtt_agent_config/core_mqtt_agent and that causing warnings that logging has been redefined. perhaps these logging macros should just be moved to the default config file instead of having two files? or if the plan is to separate config from logging then probably have an option for it similar to the default config file and customers then implement core_mqtt_agent_config.h if DO_NOT_USE_CUSTOM_CONFIG is set and core_mqtt_agent_logging.h if DO_NOT_USE_LOGGING is set?

source/include/core_mqtt_agent_config_defaults.h
AniruddhaKanhere: It is not used as an array index or a variable. It is just used to enable QoS publishes if it is non zero. I think we can leave it as is.

source/include/core_mqtt_agent_default_logging.h
AniruddhaKanhere: I did not understand properly. The logging config and the other config is completely separate. We should not mix those.
The logging configuration is supposed to be in `core_mqtt_agent_config.h` file.
The user provided configuration will override the definitions in both the lib-config and logging-config files.

So, the user only has to include one config file.

As for being redefined - it should not happen as the logging macros are guarded by ifndef

jasonpcarroll: I guess I was a bit confused as to why this file is needed in the first place - to me it feels like a potential trap because if a file includes this file before actual logging definitions you will get re-definition warnings (i.e. including this file then core_mqtt_agent.h). not a huge gripe. i do think that this should be included in core_mqtt_agent.h after core_mqtt_agent_defaults.h instead of included in core_mqtt_agent.c as it is. I'd also rec just having these default logging configs in core_mqtt_agent_config_defaults.h since that would be consistent with https://github.com/FreeRTOS/coreHTTP/blob/main/source/include/core_http_config_defaults.h.

jasonpcarroll (dismissed):

source/include/core_mqtt_agent_default_logging.h
AniruddhaKanhere: Thank you @jasonpcarroll for approving the PR. I am putting the excerpt of our conversation here for the next reader.

- The default headers are not supposed to be included by any application(s).
- Moving the includes of default header files to core_mqtt_agent.h will essentially lead to leaking of logging configuration to the other files using the Agent (like OTA).
- The core_mqtt_agent.h still has included a config file but that cannot be fixed without modifying the API as the MQTTAgent struct declares a configurable array to hold information about unacknowledged packets.

While this doesn't fix the leaking of logging macros on its own, it is a step in the right direction - which we would continue to move towards.

source/core_mqtt_agent.c
```diff
@@ -621,9 +648,9 @@ static void handleAcks( const MQTTAgentContext_t * pAgentContext,
 
 static MQTTAgentContext_t * getAgentFromMQTTContext( MQTTContext_t * pMQTTContext )
 {
-    void * ret = pMQTTContext;
+    size_t offset = ( ( size_t ) &( ( ( MQTTAgentContext_t * ) 0 )->mqttContext ) );
 
-    return ( MQTTAgentContext_t * ) ret;
+    return ( MQTTAgentContext_t * ) &pMQTTContext[ 0U - offset ];
```
archigup, line 653: Does this not have to be cast to uint8_t? Otherwise skipping offset*sizeof( MQTTContext_t * ) backwards. Also probably should not be dereferenced, as is relative to pointed to data rather than pointer itself?

source/core_mqtt_agent.c
archigup: I'm pretty sure this now invokes undefined behavior though

archigup: Also size_t is not guaranteed to hold a pointer converted to int (This is the case on at least one upstream FreeRTOS port and on CHERI archs).

archigup: Maybe
```
static MQTTAgentContext_t * getAgentFromMQTTContext( MQTTContext_t * pMQTTContext )
{
    MQTTAgentContext_t  ctx = { 0 };
    ptrdiff_t offset = ( ( uint8_t * ) &( ctx->mqttContext ) ) - ( uint8_t * ) ctx;
    return ( MQTTAgentContext_t * ) &( ( ( uint8_t * ) pMQTTContext )[ - offset ] );
}
```

jasonpcarroll (approved):

PR #81, opened 2023-01-30
Cast MQTTAgentCommandType_t to unsigned

cookpate:
## Background

The underlying integer type of an enum in C is implementation-defined. 
It may be compatible with one of: 
* char 
* unsigned integers
* signed integers

IAR Embedded Workbench decides to make the enum unsigned only to warn against checking an enum value is positive.
Other compilers may emit signed integers for the enum, so an explicit cast is needed.

### DIFF
--- a/source/core_mqtt_agent.c
+++ b/source/core_mqtt_agent.c
@@ -562,9 +562,9 @@ static MQTTStatus_t processCommand( MQTTAgentContext_t * pMqttAgentContext,
 
     if( pCommand != NULL )
     {
-        assert( pCommand->commandType < NUM_COMMANDS );
+        assert( ( unsigned int ) pCommand->commandType < ( unsigned int ) NUM_COMMANDS );
 
-        if( ( pCommand->commandType >= NONE ) && ( pCommand->commandType < NUM_COMMANDS ) )
+        if( ( unsigned int ) pCommand->commandType < ( unsigned int ) NUM_COMMANDS )
         {
             commandFunction = pCommandFunctionTable[ pCommand->commandType ];
             pCommandArgs = pCommand->pArgs;


cookpate:
Interesting, lcov/gcov check 2 extra branches when MQTTAgentCommandType_t is not an enum type. It expects a case where commandType is both less than NONE and greater than or equal to NUM_COMMANDS.

The workaround is to typedef MQTTAgentCommandType_t to an unsigned int and remove the NONE check since commandType would then always be positive.

cookpate:
> Why does the enum have to be converted to an int? Is it possible to just cast the two values at the comparison site?

There isn't a mechanism to ensure that this unsigned cast happens at every comparison site. An additional PR could arrive later to add a new use for the command type enum in a different function, and it will reintroduce that warning for IAR if the unsigned cast isn't applied.

We currently do not build with IAR in GitHub Actions, so this warning will certainly be overlooked.

source/include/core_mqtt_agent.h
```diff
@@ -58,7 +58,14 @@ typedef enum MQTTCommandType
...
     NUM_COMMANDS /**< @brief The number of command types handled by the agent. */
-} MQTTAgentCommandType_t;
+};
+
+/**
+ * @ingroup mqtt_agent_enum_types
+ * @brief use with values of enum MQTTCommandType
+ * to set the command type in struct MQTTAgentCommand
+ */
+typedef uint32_t MQTTAgentCommandType_t;
```
AniruddhaKanhere, line 68: This removes any 'hints' that an IDE gets from the type of the variable. When writing code, the smart suggestions that IDEs provide seem important to me.

Secondly, why a `uint32_t`? Why not a `uin8_t`? I am curious.

AniruddhaKanhere (approved):
Approved with comments.

I do not like making a variable which will always hold an enum value into a `uint32_t`.

source/include/core_mqtt_agent.h
cookpate: I hadn't considered IDE auto-completions. I'll revert and try casting as Archit suggested.

archigup (approved):

rawalexe (approved):

PR #85, opened 2023-02-23
Test GitHub Action with Centralized definition

karkhaz:
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/test/cbmc/proofs/lib/print_tool_versions.py
+++ b/test/cbmc/proofs/lib/print_tool_versions.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+#
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT-0
+
+
+import logging
+import pathlib
+import shutil
+import subprocess
+
+
+_TOOLS = [
+    "cadical",
+    "cbmc",
+    "cbmc-viewer",
+    "cbmc-starter-kit-update",
+    "kissat",
+    "litani",
+]
+
+
+def _format_versions(table):
+    lines = [
+        "<table>",
+        '<tr><td colspan="2" style="font-weight: bold">Tool Versions</td></tr>',
+    ]
+    for tool, version in table.items():
+        if version:
+            v_str = f'<code><pre style="margin: 0">{version}</pre></code>'
+        else:
+            v_str = '<em>not found</em>'
+        lines.append(
+            f'<tr><td style="font-weight: bold; padding-right: 1em; '
+            f'text-align: right;">{tool}:</td>'
+            f'<td>{v_str}</td></tr>')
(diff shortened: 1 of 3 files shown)


.github/workflows/proof_ci_resources/config.yaml
```diff
@@ -0,0 +1,7 @@
+cadical-tag: latest
```
paulbartell, line 1: Does this file get used by anything other than the github actions? Could these config options be passed into the github action as parameters?

test/cbmc/proofs/lib/print_tool_versions.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
```
paulbartell, line 1: Could this be part of the code in CI-CD-Github-Actions ?

test/cbmc/proofs/lib/summarize.py
```diff
@@ -1,8 +1,28 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
paulbartell, line 1: Similar question regarding this file as well.

.github/workflows/proof_ci_resources/config.yaml
karkhaz: The idea is that we'd like to be able to add new configuration parameters if we ever add more features to the proof CI Action.

We have a script to automatically update & preserve settings, but we'd rather have a separate file that contains only the settings, rather than writing over the `ci.yml` file that may be arranged in a particular way.

I'm not completely opposed to using parameters, but I think keeping them out of the ci.yml file is helpful when we want to update 14 repositories at once.

paulbartell: My preference would be to split into two separate actions:

The first action would setup the runner with the necessary prerequisites (could be replaced by a container image in the future). This action would take arguments for the following:
* cadical-tag
* cbmc-version
* kissat-tag
* litani-version

With sane defaults in the definition in the CI-CD-Github-Actions repository so most spoke repositories don't need to set anything specific.

The second action would run set of CI proofs given the following arguments:
* proofs-dir (with a reasonable default?)
* proofs (list of proofs to run or run all by default).
* script-name (if needed... I think it's probably run-cbmc-proofs.py in all of the spoke repositories right?)

test/cbmc/proofs/lib/print_tool_versions.py
paulbartell: ok.

.github/workflows/proof_ci_resources/config.yaml
paulbartell: The second action could also probably just be a normal shell-script step. Maybe thats easier?

PR #124, opened 2025-02-21
Updating the coreMQTT submodule pointer

DakshitBabbar:
Description
-----------
- Updates the coreMQTT submodule pointer to the latest commit of the library.
- Includes the `core_mqtt_config_defaults.h` file in the `core_mqtt_agent_config_defaults.h`. 
    - This is done because the older version of coreMQTT (v1.x.x) has `core_mqtt_config.h` included in the `core_mqtt.h` file whose macros is used by a couple of more files like `freertos_command_pool.c` in some old demos (like the old OTA demos). But in the new version this includes is removed from the `core_mqtt.h` file and added in the `core_mqtt_config_defaults.h` file. A simple solution to this problem was to add the `core_mqtt_config_defaults.h` in the `core_mqtt_agent_config_defaults.h` which is included in the `core_mqtt_agent.h` that is used by the files that require the macros. 
- Updates README for correct commands to build unit tests
- Modifies the cmake files for building the unit tests successfully on mac machines
- Modifies the unit tests for full coverage (initially coverage was different for mac and linux)

Test Steps
-----------
- Verified the updates with the demos in the [core-mqtt-agent-demos](https://github.com/FreeRTOS/coreMQTT-Agent-Demos) repository.

Checklist:
----------

- [x] I have tested my changes. No regression in existing tests.
- [x] I have modified and/or added unit-tests to cover the code changes in this Pull Request.

Related Issue
-----------

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/tools/cmock/create_test.cmake
+++ b/tools/cmock/create_test.cmake
@@ -129,7 +129,7 @@ function(create_mock_list mock_name
                                ${mocks_dir}
                                ${mock_include_list}
            )
-           if (APPLE)
+        if (APPLE)
            set_target_properties(${mock_name} PROPERTIES
                    LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib
                    POSITION_INDEPENDENT_CODE ON
--- a/tools/cmock/create_test.cmake
+++ b/tools/cmock/create_test.cmake
@@ -45,7 +45,7 @@ function(create_test test_name
         add_dependencies(${test_name} ${dependency})
         target_link_libraries(${test_name} ${dependency})
     endforeach()
-    target_link_libraries(${test_name} -lgcov unity)
+    target_link_libraries(${test_name} unity)
     target_link_directories(${test_name}  PUBLIC
                             ${CMAKE_CURRENT_BINARY_DIR}/lib
             )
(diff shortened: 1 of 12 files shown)


tools/cmock/project.yml
```diff
@@ -21,6 +21,7 @@
   :includes:        # This will add these includes to each mock.
     - <stdbool.h>
     - <stdint.h>
+  :includes_c_post_header:
+    - <cmock_opaque_types.h>
   :treat_externs: :exclude  # Now the extern-ed functions will be mocked.
-  :weak: __attribute__((weak))
```
tony-josi-aws, line 25: Why is it removed?

DakshitBabbar: With the weak symbols on, on mac, the function calls to these functions go to the actual definitions instead of the mocked definitions (for the unit tests). This is because the darwin compiler picks up the first strong symbol that it finds during the linking phase but in case of linux it just takes the first symbol that it finds (regardles of weak/strong). Hence the tests were working fine on linux but not on mac.

tony-josi-aws: Is it working now on Linux?

tony-josi-aws (approved):

tools/cmock/create_test.cmake
```diff
@@ -129,11 +129,20 @@ function(create_mock_list mock_name
                                ${mocks_dir}
                                ${mock_include_list}
            )
-    set_target_properties(${mock_name} PROPERTIES
-                        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/lib
-                        POSITION_INDEPENDENT_CODE ON
-            )
-    target_compile_definitions(${mock_name} PUBLIC
+           if (APPLE)
```
aggarg, line 132:
```suggestion
        if (APPLE)
```

aggarg (approved):

tony-josi-aws (approved):

PR #140, opened 2026-06-12
Fix coreMQTT v5 API signatures in doxygen examples

### LINKED ISSUE #139: Error in Documentation
Affected Library Version: v2.0.0
[Documentation](https://freertos.github.io/coreMQTT-Agent/v2.0.0/mqtt_agent_command_function.html)
Image

since coreMQTT v5.0.0 MQTT_Disconnect requires two more Arguments.

archigup:
Fixes #139

The doxygen examples in `core_mqtt_agent.h` called coreMQTT APIs with pre-v5 signatures, which no longer compile against the pinned coreMQTT v5.0.2 submodule.

- `MQTTAgent_CommandLoop` example: `MQTT_Disconnect` now takes property-builder and reason-code arguments (passing `NULL, NULL`).
- `MQTTAgent_ResumeSession` example: `MQTT_Connect` now takes property-builder and will-property-builder arguments (passing `NULL, NULL`); also added the missing trailing semicolon.

The `MQTT_Disconnect` case is the one reported in #139. The `MQTT_Connect` example was found while sweeping the other coreMQTT API examples. Implementation code already uses the correct v5 signatures, so this is a documentation-only change.

Closes #139

### DIFF
--- a/source/include/core_mqtt_agent.h
+++ b/source/include/core_mqtt_agent.h
@@ -351,7 +351,7 @@ MQTTStatus_t MQTTAgent_Init( MQTTAgentContext_t * pMqttAgentContext,
  * {
  *    // Terminate command was processed but MQTT connection was not
  *    // closed. Thus, need to close both MQTT and socket connections.
- *    status = MQTT_Disconnect( &( mqttAgentContext.mqttContext ) );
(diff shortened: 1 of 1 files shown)


aggarg (approved):

kstribrnAmzn (approved):

kstribrnAmzn:
Link does actually exist. Not going to waste time with retrying the action.

